Issue #40, opened 2024-03-16
doc_string which is the first argument of example, must be string object


eisukeyeongjo:
### doc_string which is the first argument of example, must be string object

RSpec example recognize the first argument as the description.
But when even symbol or hash object which set intended as metadata is given in the first argument, it is recognized as the description and metadata does not work.
I want to add some change to keep doc_string a string object.

### Your environment
* Ruby version: 3.3.0
* rspec-core version: 3.1.3

### Steps to reproduce
**prepare rspec test below**

```
$ vi test.rb
$ cat test.rb
```

```ruby
context do
  it :pending do
    expect(true).to eq false
  end

  it pending: 'some reason' do
    expect(true).to eq false   
  end
end
```

### Actual behavior
metadata is recognized as description and pending does not work, failing two examples

```
$ bundle exec rspec  test.rb --format documentation

  pending (FAILED - 1)
  {:pending=>"some reason"} (FAILED - 2)

Failures:

  1) pending
     Failure/Error: expect(true).to eq false
     
       expected: false
            got: true
     
       (compared using ==)
     
       Diff:
       @@ -1 +1 @@
       -false
       +true
       
     # ./test.rb:14:in `block (2 levels) in <top (required)>'

  2) {:pending=>"some reason"}
     Failure/Error: expect(true).to eq false
     
       expected: false
            got: true
     
       (compared using ==)
     
       Diff:
       @@ -1 +1 @@
       -false
       +true
       
     # ./test.rb:18:in `block (2 levels) in <top (required)>'

Finished in 0.01336 seconds (files took 0.08613 seconds to load)
2 examples, 2 failures

Failed examples:

rspec ./test.rb:13 # pending
rspec ./test.rb:17 # {:pending=>"some reason"}
```

### Expected behavior
To prevent metadata as first argument which does not work properly, it raises ArgumentError like below when execute each example above whose first argument is not nil or string object.

```
ArgumentError:
  `:pending` is not acceptable for doc_string.
  it must be a string.
```

### Note
First, I created PR rspec/rspec-core#3074 not to set doc_string when first argument is metadata.
But I think that keeping the first argument of example doc_string is much more appropriate.
(In the first place, as commented [here](https://github.com/rspec/rspec-core/blob/main/lib/rspec/core/example_group.rb#L114)  it's the specification.)

### Related Pull Request
I proposed this new specification in the PR below,
After the discussion, we decided to add this in upcoming 4.x and add warning in 3.99

[...]

PR #107, opened 2024-11-29
[rspec-core] Add deprecate warning when example doc string is not a string object

JonRowe:
Preparing for upcoming new specification mentioned rspec/rspec#40 , added deprecated warning when example doc string is not a string object

4.0 counterpart https://github.com/rspec/rspec/pull/252 (or #339)

### What is changed

When execute rspec below

```ruby
context do
  it :pending do
    # Only pending option without reason
    expect(true).to eq false
  end

  it pending: 'only pending option' do
    expect(true).to eq false   
  end  

  it 'description with option', pending: 'some reason' do
    expect(true).to eq false
  end
end
```

Output deprecate warning like

```
Deprecation Warnings:

Hash object `{:pending=>"only pending option"}` as example doc string is deprecated. Use a string instead. Called from /home/username/test.rb:7:in `block in <top (required)>'.

Symbol object `:pending` as example doc string is deprecated. Use a string instead. Called from /home/username/test.rb:2:in `block in <top (required)>'.
```

<details>
<summary>Entire output is here</summary>

```
$ bundle exec rspec test.rb --format documentation


  pending (FAILED - 1)
  {:pending=>"only pending option"} (FAILED - 2)
  description with option (PENDING: some reason)

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) description with option
     # some reason
     
       expected: false
            got: true
     
       (compared using ==)
     
       Diff:
       @@ -1 +1 @@
       -false
       +true
       
     # ./test.rb:12:in `block (2 levels) in <top (required)>'

Failures:

  1) pending
     
       expected: false
            got: true
     
       (compared using ==)
     
       Diff:
       @@ -1 +1 @@
       -false
       +true
       
     # ./test.rb:4:in `block (2 levels) in <top (required)>'

  2) {:pending=>"only pending option"}
     
       expected: false
            got: true
     
       (compared using ==)
     
       Diff:
       @@ -1 +1 @@
       -false
       +true
       
     # ./test.rb:8:in `block (2 levels) in <top (required)>'

Deprecation Warnings:

Hash object `{:pending=>"only pending option"}` as example doc string is deprecated. Use a string instead. Called from /home/username/test.rb:7:in `block in <top (required)>'.

Symbol object `:pending` as example doc string is deprecated. Use a string instead. Called from /home/username/test.rb:2:in `block in <top (required)>'.


If you need more of the backtrace for any of these deprecations to
identify where to make the necessary changes, you can configure
`config.raise_errors_for_deprecations!`, and it will turn the
deprecation warnings into errors, giving you the full backtrace.

2 deprecation warnings total

Finished in 0.01384 seconds (files took 0.08949 seconds to load)
3 examples, 2 failures, 1 pending

Failed examples:

rspec ./test.rb:2 # pending
rspec ./test.rb:7 # {:pending=>"only pending option"}

```

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,25 +21,25 @@ jobs:
     uses: rspec/rspec/.github/workflows/rubocop.yml@main
 
   core:
-    uses: rspec/rspec/.github/workflows/rspec.yml@3-13-maintenance
+    uses: rspec/rspec/.github/workflows/rspec.yml@3-99-maintenance
     with:
       library: 'rspec-core'
-      rspec_version: '~> 3.13.0'
+      rspec_version: '~> 3.99.0'
 
   mocks:
-    uses: rspec/rspec/.github/workflows/rspec.yml@3-13-maintenance
+    uses: rspec/rspec/.github/workflows/rspec.yml@3-99-maintenance
     with:
       library: 'rspec-mocks'
-      rspec_version: '~> 3.13.0'
+      rspec_version: '~> 3.99.0'
 
   expectations:
-    uses: rspec/rspec/.github/workflows/rspec.yml@3-13-maintenance
+    uses: rspec/rspec/.github/workflows/rspec.yml@3-99-maintenance
     with:
       library: 'rspec-expectations'
-      rspec_version: '~> 3.13.0'
+      rspec_version: '~> 3.99.0'
 
   support:
-    uses: rspec/rspec/.github/workflows/rspec.yml@3-13-maintenance
+    uses: rspec/rspec/.github/workflows/rspec.yml@3-99-maintenance
     with:
       library: 'rspec-support'
-      rspec_version: '~> 3.13.0'
+      rspec_version: '~> 3.99.0'
--- a/rspec-core/Changelog.md
+++ b/rspec-core/Changelog.md
@@ -3,7 +3,7 @@
 
 Deprecations:
 
-* Add RSpec 4 deprecation warnings. (Phil Pirozhkov, Jon Rowe, rspec/rspec-core#2880, rspec/rspec#106)
+* Add RSpec 4 deprecation warnings. (Phil Pirozhkov, Jon Rowe, rspec/rspec-core#2880, rspec/rspec#106, rspec/rspec#243)
 
 Enhancements:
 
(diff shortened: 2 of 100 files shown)


pirj:
Last one for 3.99 is green.

However, this does not cover all cases where #252 raises, it only warns for improper example descriptions.

This brings me back to the `described_thing` discussion (that I fail to find quickly). I'm against putting an object as the first argument just to be able to skip the `subject { that_object }` line. Am I missing some obvious cases where this makes an intuitively readable spec?

If not, I intend to adjust those warnings to match exceptions raised in #252.

pirj:
This should be green this time, too

pirj:
@JonRowe Just a heads-up. GitHub closed this, apparently due to this:

image

JonRowe:
Yeah I accidentally deleted the branch when rebasing (its not protected because its essentially a dev branch until theres a release or beta) need to check there aren't any more closed PRs, I'll probably setup a branch rule now :joy:

pirj:
Also failed the same way in #243

pirj:
@JonRowe how do you feel about making this a configurable opt-in?
```ruby
RSpec.configure do |config|
  config.strict_docstring! # enforce the use of string as the very first argument
end
```

This way, we can make it the default in the project initializer’s spec_helper.rb in RSpec 4.0
We can keep this warning in 3.99, but add a note that it can be ignored/disabled if it’s overkill for users to fix all their existing specs.

In 4.0, unless strict is set, we’ll still consider the first argument metadata if it’s a symbol/hash, and won’t coerce into a string. If it’s a numeric/object/array etc, we’ll `to_s`/`inspect` it into a string, effectively keeping the RSpec 3 behaviour.
If strict is set, and the first argument is present and is not a string, we’ll quit with an error.

I’ll be happy to work on this for both 3.99 and 4.0.

We can also completely leave this out, and create a separate gem that augments RSpec’s behaviour in the same way. I can work on this, too.


Unless I’m missing something, this is the last bit remaining for RSpec 4?


It would be nice to synchronise the RSpec 4 release with the Ruby 4 release.


I’ve played a bit with removing the `ruby2_keywords` for the `*, **` style delegation, but haven’t dragged it to completion. Can’t tell if we would have to drop support for Ruby 3.1 to be able to do this, I couldn’t find sufficient confirmation for this in Benoit Daloze’s blog posts, which we used as a guideline before.
Anyway, it’s a different topic, and can be done in 4.1.

petrenkorf:
No updates on this?

PR #116, opened 2024-11-30
[rspec-mocks] Prevent error formatting from causing recursive errors

JonRowe:
This is rspec/rspec-mocks#1275


JonRowe:
Released as 3.13.3

PR #126, opened 2024-11-30
[rspec-expectations] Update StartWith, EndWith to use native methods if available

JonRowe:
This is rspec/rspec-expectations#1327

> This addresses issue #1025.
>
> With this change, the StartWith matcher will rely on an object's start_with? method if available.  Similarly, the EndWith matcher will rely on an object's end_with? method if available.
>
> This is especially useful when a class implements start_with? but not the indexing operator, or end_with? but not the indexing operator.


JonRowe:
I've migrated this to not loose track of it, but I feel this needs to "fall back" to this behaviour rather than override

PR #152, opened 2024-12-02
Fix source_code_uri

mopp:
## Overview

The URL of the gemspecs is old.
This PR replaces them.


JonRowe:
Thanks, we're still in the progress of completing the migration but it doesn't hurt to have this changed now.

Issue #153, opened 2024-12-02
Monorepo migration progress


JonRowe:
# Monorepo migration progress

- [x] Combine repo history
- [x] Rebuild CI for combined repo
- [x] Transfer issues
- [x] Migrate PRs
- [x] Setup build for 3-13-maintenance for patch releases
- [x] Switch to main
- [ ] Migrate documentation setup from rspec-dev
- [x] Sort out old tags
- [x] Handle change logs
- [x] Refactor gem-specs to point links to this repo:
  - [x] rspec-core
  - [x] rspec-expectations
  - [x] rspec-mocks
  - [x] rspec-support
- [x] Check for broken links
- [x] Finalise CI for 4-0 development
  - [x] Add windows for at least rspec-core
  - [x] Add one "stable" JRuby build

~- [ ] Add rspec-rails cross build?~

JonRowe:
For tags the latest version is tagged for each gem, for > 4.0.0 we may change towards a unified point release of all gems at the same time e.g. how Rails does it, but for now this replicates out 3.13.x development flow, and I've updated the change logs for the latest diffs for the current changes, the historical links will remain and I think in 4.0.0 we will truncate them linking to these historical ones.

JonRowe:
For documentation we need to bring across and simplify the Rake task from rspec-dev, I'm also wondering if we could host just the docs from this repo, meaning potentially we could setup a github action to publish them...

pirj:
@JonRowe I have capacity to work on this.

> Refactor gem-specs to point links to this repo

Is there anything left to be done? At a glance, gemspecs are fine.

> Check for broken links

https://github.com/rspec/rspec/pull/244

>  Finalise CI for 4-0 development 

https://github.com/rspec/rspec/pull/245

JonRowe:
Realistically the last thing remaining is to move documentation building over here properly. I've done some experiments with autobuilding docs but nothing complete enough to PR.

PR #154, opened 2024-12-04
Fix bug_tracker_uri

yskttm:
rspec-metagem has been archived, so replace it with the new rspec URL


benoittgt (approved):

JonRowe:
Thanks theres more to do here, we need to update all the gem-specs like rspec-support has then backport to 3-13-maintenance so patch releases pick up this changes (might even be worth doing a patch release to update the links), I've added this to the (rspec/rspec#153) issue to track

PR #155, opened 2024-12-04
Status badge report wrong branch

benoittgt:
Hello

Congrats for the monorepo. Happy to see it. <3

I noticed that the badge report was in failing state even if the CI on main was green. I think we should scope to `ci.yml` and main branch.


JonRowe:
Thanks, I've added a note to the (rspec/rspec#153) issue to check for more

PR #156, opened 2024-12-04
Use a thread based invoking check

JonRowe:
This is a follow up to #116 which introduced a threaded race condition for JRuby, this changes the invoking check to being thread local based on the object id


JonRowe:
Released as 3.13.3

PR #160, opened 2024-12-13
Add windows back to CI

JonRowe:
We should run at least one windows build, theres two here only because of my uncertainty of 3.3 support where as I know 2.7 worked.


JonRowe:
Merged in 3b8dfa8b6fbffb0cb45effe79c072b937af325b9

PR #161, opened 2024-12-17
Set `:all`/`:deprecation_only`/`:none` as options for `config.warnings=`

### LINKED ISSUE #37: Warning[:deprecated] = true by default in RSpec
### Subject of the issue

See https://bugs.ruby-lang.org/issues/17591.

The plan of ruby-core hiding deprecation warnings in Ruby 2.7.2 is that test frameworks, like rspec, should enable them by default (with `Warning[:deprecated] = true if Warning.respond_to?(:[]=)`, which only enables deprecation warnings).
This is important so that developers see warnings in development, and that they see the warnings before updating to the next Ruby version.

Of course, RSpec users can still choose to disable deprecation warnings explicitly if they really want with `Warning[:deprecated] = false` (but then they should not be surprised when it breaks hard when updating to a newer Ruby, that's the cost of ignoring deprecation warnings).

Somewhat related to https://github.com/rspec/rspec-core/issues/2809.

### Your environment
* Ruby version: 2.7.2
* rspec-core version: 3.10.1

### Steps to reproduce
```ruby
# some_spec.rb 
def foo(**kwargs)
  kwargs
end

RSpec.describe 'Test' do
  it "should warn about keyword arguments" do
    expect(foo({a: 1})).to eq({a: 1})
  end
end
```

`rspec some_spec.rb`

### Expected behavior
```
$ rspec some_spec.rb
/home/eregon/code/rspec-ws/some_spec.rb:7: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/eregon/code/rspec-ws/some_spec.rb:1: warning: The called method `foo' is defined here
.

Finished in 0.00154 seconds (files took 0.05064 seconds to load)
1 example, 0 failures
```
### Actual behavior
No warnings in 2.7:
```
$ rspec some_spec.rb
.

Finished in 0.00157 seconds (files took 0.05947 seconds to load)
1 example, 0 failures
```

Breaks hard in 3.0:
```
$ rspec some_spec.rb
F

Failures:

  1) Test should warn about keyword arguments
     Failure/Error:
       def foo(**kwargs)
         kwargs
       end
     
     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # ./some_spec.rb:1:in `foo'
     # ./some_spec.rb:7:in `block (2 levels) in <top (required)>'

Finished in 0.00121 seconds (files took 0.05637 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./some_spec.rb:6 # Test should warn about keyword arguments
```

byroot:
Fix: https://github.com/rspec/rspec/issues/37

Since Ruby 2.7.2, Ruby expect test frameworks to enable deprecation warnings so users get advance notice of future breaking changes.

I implemented it as specified in https://github.com/rspec/rspec/issues/37#issuecomment-2504702395

3.99 deprecation for boolean `config.warnings = true`/`config.warnings = false` https://github.com/rspec/rspec/pull/268

### DIFF
--- a/rspec-core/lib/rspec/core/configuration.rb
+++ b/rspec-core/lib/rspec/core/configuration.rb
@@ -489,6 +489,15 @@ module RSpec
         @world = World::Null
         @pending_failure_output = :full
         @force_line_number_for_spec_rerun = false
+        @warnings = nil
+
+        if warnings?
+          # Turn on all warnings if warnings already configured
+          self.warnings = :all
+        else
+          # Turn on Ruby deprecations by default
+          self.warnings = :deprecations_only
+        end
 
         define_built_in_hooks
       end
@@ -1633,10 +1642,36 @@ module RSpec
       # @private
       delegate_to_ordering_manager :seed_used?, :ordering_registry
 
-      # Set Ruby warnings on or off.
+      # Set Ruby warnings. `:deprecations_only` is recommended, but may be too
+      # noisy due to dependencies.
+      # Can be set to `:all` or `:none` to show all warnings or none of them.
       def warnings=(value)
-        $VERBOSE = !!value
+        case value
+        when :none
+          @warnings = :none
+          $VERBOSE = false
+          ::Warning[:deprecated] = false
+        when :all
+          @warnings = :all
+          $VERBOSE = true
+          ::Warning[:deprecated] = true
+        when :deprecations_only
+          @warnings = :deprecations_only
+          $VERBOSE = false
+          ::Warning[:deprecated] = true
+        else
+          raise "Unsupported value for `warnings` (#{value.inspect}). " \
+                "Only `:none`, `:all` and `:deprecations_only` are supported."
+        end
       end
+
+      # @return [Boolean] Whether or not ruby deprecations are enabled.
+      def deprecation_warnings?
+        ::Warning[:deprecated]
+      end
+
+      # @return [Symbol] The configured warning level
+      attr_reader :warnings
 
       # @return [Boolean] Whether or not ruby warnings are enabled.
       def warnings?
--- a/rspec-core/lib/rspec/core/option_parser.rb
+++ b/rspec-core/lib/rspec/core/option_parser.rb
@@ -178,11 +178,7 @@ module RSpec::Core
         end
 
         parser.on('-w', '--warnings', 'Enable ruby warnings') do
-          if Object.const_defined?(:Warning) && Warning.respond_to?(:[]=)
-            # :nocov: on older Ruby without Warning
-            Warning[:deprecated] = true
-            # :nocov:
-          end
+          Warning[:deprecated] = true
           $VERBOSE = true
         end
 
--- a/rspec-core/lib/rspec/core/project_initializer/spec/spec_helper.rb
+++ b/rspec-core/lib/rspec/core/project_initializer/spec/spec_helper.rb
@@ -35,6 +35,11 @@ RSpec.configure do |config|
 # The settings below are suggested to provide a good initial experience
 # with RSpec, but feel free to customize to your heart's content.
 =begin
+  # This setting turns on all Ruby warnings. It's recommended, but in some cases may
+  # be too noisy due to issues in dependencies. The default is set to `:deprecations_only`
+  # but it can also be set to `:none` to suppress them.
+  config.warnings = :all
+
   # This allows you to limit a spec run to individual examples or groups
   # you care about by tagging them with `:focus` metadata. When nothing
   # is tagged with `:focus`, all examples get run. RSpec also provides
@@ -46,10 +51,6 @@ RSpec.configure do |config|
   # the `--only-failures` and `--next-failure` CLI options. We recommend
   # you configure your source control system to ignore this file.
   config.example_status_persistence_file_path = "spec/examples.txt"
-
-  # This setting enables warnings. It's recommended, but in some cases may
-  # be too noisy due to issues in dependencies.
-  config.warnings = true
 
   # Many RSpec users commonly either run the entire suite or an individual
   # file, and it's useful to allow more verbose output when running an
--- a/rspec-core/spec/rspec/core/configuration_spec.rb
+++ b/rspec-core/spec/rspec/core/configuration_spec.rb
@@ -2501,27 +2501,63 @@ module RSpec::Core
 
     describe '#warnings' do
       around do |example|
+        original_deprecated = ::Warning[:deprecated]
         original_setting = $VERBOSE
         example.run
         $VERBOSE = original_setting
+        ::Warning[:deprecated] = original_deprecated
       end
 
-      it "sets verbose to true when true" do
-        config.warnings = true
+      it "is configured to deprecations_only by default" do
+        $VERBOSE = false
+        blank_config = Configuration.new
+        expect(blank_config.warnings).to eq :deprecations_only
+      end
+
+      it "preserves outside warning settings" do
+        $VERBOSE = true
+        blank_config = Configuration.new
+        expect(blank_config.warnings).to eq :all
+      end
+
+      it "sets $VERBOSE and Warning[:deprecated] to true when :all" do
+        config.warnings = :all
         expect($VERBOSE).to eq true
+        expect(::Warning[:deprecated]).to eq true
       end
 
-      it "sets verbose to false when true" do
-        config.warnings = false
+      it "sets $VERBOSE and Warning[:deprecated] to false when :none" do
+        config.warnings = :none
         expect($VERBOSE).to eq false
+        expect(::Warning[:deprecated]).to eq false
       end
 
-      it 'returns the verbosity setting' do
-        config.warnings = true
-        expect(config.warnings?).to eq true
+      it "sets $VERBOSE to false but Warning[:deprecated] to true when :deprecations_only" do
+        config.warnings = :deprecations_only
+        expect($VERBOSE).to eq false
+        expect(::Warning[:deprecated]).to eq true
+      end
 
-        config.warnings = false
+      it "raises on unsupported warnings value" do
+        [:unknown, true, false].each do |unacceptable|
+          expect {
+            config.warnings = unacceptable
+          }.to raise_error(a_string_including("Unsupported value for `warnings` (#{unacceptable.inspect})"))
+        end
+      end
+
+      it 'returns the verbosity settings' do
+        config.warnings = :all
+        expect(config.warnings?).to eq true
+        expect(config.deprecation_warnings?).to eq true
+
+        config.warnings = :deprecations_only
         expect(config.warnings?).to eq false
+        expect(config.deprecation_warnings?).to eq true
+
+        config.warnings = :none
+        expect(config.warnings?).to eq false
+        expect(config.deprecation_warnings?).to eq false
       end
 
       it 'is loaded from config by #force' do
--- a/rspec-core/Changelog.md
+++ b/rspec-core/Changelog.md
@@ -35,12 +35,15 @@ Breaking Changes:
   (Phil Pirozhkov, rspec/rspec-core#2864)
 * Unify multi-condition filtering to use "all" semantic. (Phil Pirozhkov, rspec/rspec-core#2874)
 * Change the default order to random. (Santiago Bartesaghi, rspec/rspec-core#2929)
+* Default warning level is now `:deprecations_only`. (Jon Rowe, rspec/rspec#161)
 
 Enhancements:
 
 * Add config option (`RSpec::Core::Configuration#force_line_number_for_spec_rerun`) to
   allways print a line number rather than an example id when the line number is ambiguous.
   (Baden Ashford, rspec/rspec-core#3085)
+* Expand `RSpec::Core::Configuration#warnings=` to take `:all`, `:deprecations_only` and `:none`
+  options. (Jean Boussier, rspec/rspec#161)
 
 ### 3.13.4 / 2025-05-27
 [Full Changelog](http://github.com/rspec/rspec/compare/rspec-core-v3.13.3...rspec-core-v3.13.4)
--- a/rspec-support/lib/rspec/support/spec.rb
+++ b/rspec-support/lib/rspec/support/spec.rb
@@ -19,6 +19,8 @@ RSpec.configure do |c|
   c.include RSpec::Support::FormattingSupport
   c.include RSpec::Support::InSubProcess
 
+  c.warnings = :all
+
   unless defined?(Debugger) # debugger causes warnings when used
     c.before do
       warning_preventer.reset!


byroot:
Hum, CI fails because of the 100% coverage target, but I don't know how I can satisfy this given some code path are only ever entered by some specific Ruby versions.

I'll try to do some static define instead but I don't think it will work.

byroot:
Alright, I tried a few workarounds but they're all awful and don't pass the coverage check anyway, so I'll wait for instruction from the maintainers.

rspec-core/lib/rspec/core/configuration.rb
```diff
@@ -1773,6 +1773,34 @@ def warnings?
...
+      def deprecation_warnings?
+        if defined?(::Warning) && ::Warning.respond_to?(:[]=)
+          ::Warning[:deprecated]
+        else
+          false
+        end
+      end
+
+      # @private
+      def deprecation_warnings_set?
```
pirj: Would [:nocov:](https://github.com/simplecov-ruby/simplecov?tab=readme-ov-file#ignoringskipping-code) help here?

byroot: Ah TIL, thank you!

byroot:
Alright, I believe the remaining errors are unrelated to my changes?

rspec-core/lib/rspec/core/configuration.rb
```diff
@@ -1773,6 +1773,40 @@ def warnings?
...
+        end
+
+        # @return [Boolean] Whether or not ruby warnings are enabled.
+        def deprecation_warnings?
+          ::Warning[:deprecated]
+        end
+
+        # @private
+        def deprecation_warnings_set?
+          @deprecation_warnings_set || ::Warning[:deprecated]
```
JonRowe: This is causing a warning because if it hasn't been set to true its not initialised. As we check our code is warning free this fails the build (there is also a flakey bisect test but I can restart that when it rears its head)

rspec-core/lib/rspec/core/configuration.rb
```diff
@@ -1773,6 +1773,40 @@ def warnings?
         $VERBOSE
       end
 
+      if defined?(::Warning) && ::Warning.respond_to?(:[]=)
+        # Set Ruby deprecation warnings on or off.
+        def deprecation_warnings=(value)
+          @deprecation_warnings_set = true
+          ::Warning[:deprecated] = value
```
JonRowe: Personally, I'd rather we overload `warnings=` here to be one of: `false` `:deprecations_only` `:all` or `true` where true has the same effect as `:all`

byroot: I suppose that make sense, one downside I can think of however if that since regular warnings are quite verbose, it's likely that a bunch of projects have `warnings = false` and won't ever know about deprecation warnings.

So while I agree making it a difference config is a bit less clean, it give every RSpec user the opportunity to know about it.

But that's your call.

rspec-core/lib/rspec/core/configuration.rb
byroot: Ah right,, that's indeed an issue until Ruby 2.7. Fied it, thanks.

rspec-core/lib/rspec/core/configuration.rb
JonRowe: I think that risk is low as false was the default, but I propose that the setting be `:none`, `:deprecations_only` and `:all` with `true`/`false` synonyms for `:all`/`:none` that in turn issue an RSpec deprecation warning asking people to update to the specifics.

byroot: Ah I see. Ok, I'll try to implement that then.

byroot: Alright, I think I implemented what you suggested. I'm not really familiar with RSpec codebase though, so hopefully it's implemented the proper way?

benoittgt:
Hello Jean

I see two failures in feature specs.

```
bundle exec cucumber features/configuration/enable_global_dsl.feature:40
bundle exec cucumber features/command_line/ruby.feature:10
```
It happens in the CI after the simplecove coverage tests...

The issue seems related to how we try to report caller lines when using `.warning`. Maybe @JonRowe knows if `.warning` should be used here. I'm tempted to change how we report caller line. In my test I have 14 frames and with the current setup we are getting outside of frames before looping 3 times. 😄

Quick and dirty.

```diff
--- a/rspec-support/lib/rspec/support/caller_filter.rb
+++ b/rspec-support/lib/rspec/support/caller_filter.rb
@@ -65,7 +65,7 @@ module RSpec
 
         loop do
           stack = caller_locations(skip_frames, increment)
-          raise "No non-lib lines in stack" unless stack
+          break unless stack
```

Snippet to reproduce.
```ruby
require "bundler/inline"
gemfile(true) do
  source "https://rubygems.org"
  gem 'rspec', path: '../rspec'
end

require 'rspec/autorun'

RSpec.describe 'Joyeux Noël' do
  it do
    expect({ :a => 1, :b => 2 }).not_to include(:a, :b)
  end
end
```

rspec-core/lib/rspec/core/runner.rb
```diff
@@ -130,6 +130,12 @@ def configure(err, out)
         @configuration.error_stream = err
         @configuration.output_stream = out if @configuration.output_stream == $stdout
         @options.configure(@configuration)
+
+        unless @configuration.warnings_set?
+          RSpec.warning "config.warnings isn't set. Please set it to either `:all`, `:deprecations_only?`, or " \
```
eregon:
```suggestion
          RSpec.warning "config.warnings isn't set. Please set it to either `:all`, `:deprecations_only`, or " \
```

rspec-core/lib/rspec/core/runner.rb
```diff
@@ -130,6 +130,12 @@ def configure(err, out)
...
+                        "`:none`. Setting it to at least `:deprecations_only?` is highly" \
```
eregon:
```suggestion
                        "`:none`. Setting it to at least `:deprecations_only` is highly" \
```

byroot:
I fixed the typos in the message.

For the remaining failures, I don't know what to do, so I'll await further instructions.

rspec-core/spec/rspec/core/configuration_spec.rb
```diff
@@ -2691,34 +2691,58 @@ def strategy.order(list)
...
-    describe '#warnings' do
-      around do |example|
-        original_setting = $VERBOSE
-        example.run
-        $VERBOSE = original_setting
-      end
+    if defined?(::Warning) && ::Warning.respond_to?(:[]=)
+      describe '#warnings' do
+        around do |example|
+          original_setting = ::Warning[:deprecated]
```
pirj:
```suggestion
```
Detected by RuboCop as "Useless assignment to variable - original_setting.".

byroot: Indeed, silly mistake. I fixed it. thanks.

rspec-core/lib/rspec/core/configuration.rb
```diff
@@ -1765,14 +1766,57 @@ def self.delegate_to_ordering_manager(*methods)
...
       def warnings=(value)
-        $VERBOSE = !!value
+        @warnings_set = true
+
+        case value
+        when true, false
+          RSpec.deprecate(
+            "RSpec::Core::Configuration#warnings=",
+            :message =>  "config.warnings = #{value} is deprecated, please set warnings to one of: " \
+            "`:none`, `:all` or `:deprecations_only.`"
```
pirj:
```suggestion
            :message =>  "config.warnings = #{value} is deprecated, please set warnings to one of: " \
                         "`:none`, `:all` or `:deprecations_only.`"
```

RuboCop suggests aligning the indentation.

rspec-core/lib/rspec/core/configuration.rb
```diff
@@ -1765,14 +1766,57 @@ def self.delegate_to_ordering_manager(*methods)
...
+          $VERBOSE = false
+          self.deprecation_warnings = false
+        when :all
+          $VERBOSE = true
+          self.deprecation_warnings = true
+        when :deprecations_only
+          $VERBOSE = false
+          self.deprecation_warnings = true
+        else
+          raise "Unsupported value for `warnings` (#{value.inspect}). " \
```
pirj, line 1663: It's this line that decreases coverage (from an unreal 100%)

image

byroot: Good catch, I added a test for it.

byroot:
The remaining failures are what @benoittgt pointed out. 

Both failing scenarios are related to `rspec/autorun`:

```
Failing Scenarios:
cucumber features/command_line/ruby.feature:10 # Scenario: Require `rspec/autorun` from a spec file
cucumber features/configuration/enable_global_dsl.feature:40 # Scenario: By default rspec/autorun allows the DSL to be used globally
```

It fails because it's emitting a warning about `config.warnings` not being set.

rspec-core/lib/rspec/core/runner.rb
```diff
@@ -130,6 +130,12 @@ def configure(err, out)
         @configuration.error_stream = err
         @configuration.output_stream = out if @configuration.output_stream == $stdout
         @options.configure(@configuration)
+
+        unless @configuration.warnings_set?
+          RSpec.warning "config.warnings isn't set. Please set it to either `:all`, `:deprecations_only`, or " \
+                        "`:none`. Setting it to at least `:deprecations_only` is highly" \
+                        "recommended to receive advance notice of future Ruby breaking changes."
```
pirj:
```suggestion
                        "recommended to receive advance notice of future Ruby breaking changes.",
                        :call_site => nil
```

This was a bit tricky.
So, when we run our cucumber tests (/features), specifically those two that `require "rspec/autorun"` issue the warning that "config.warnings isn't set".
It then attempts to find a first line where this was triggered outside RSpec's own code, and fails.
This fixed `cucumber` in rspec-core for me locally.
Sorry for the inconvenience, and thanks for the new feautre!

byroot:
Thanks for the fix, I was a bit stuck because of my limited knowledge of rspec.

One of the spec I added now complains but I can fix that.

rspec-core/spec/rspec/core/runner_spec.rb
```diff
@@ -363,6 +363,19 @@ def interrupt
...
       end
 
+      it "emits a warning if config.deprecation_warnings isn't set" do
+        allow(RSpec).to receive(:deprecate)  # remove this and should_receive when ordered
+        stdout = StringIO.new
+        allow(config).to receive(:load_spec_files)
+        allow(config).to receive(:reporter).and_return(double.as_null_object)
+
+        expect(config).to receive(:warnings_set?).and_return(false)
+        expect(RSpec).to receive(:warning).with(/config\.warnings isn't set/)
```
pirj:
```suggestion
        expect(RSpec).to receive(:warning).with(/config\.warnings isn't set/, :call_site => nil)
```

pirj:
~That one seemed flaky and caused by order dependence (/tmp/aruba isn't getting created when `rspec spec/integration/order_spec.rb is run). But the seemingly trivial fix, mocking other similar specs that use Aruba, didn't work. I can reproduce this locally.~ PS fixed. PPS merged in #176

I don't mind reverting my last commit and merging this PR as is, as this failure seems largely unrelated to this PR. @JonRowe WDYT?
~But if it's a quick fix, I hope you don't mind including it.~ I can extract the flaky fix.

My apologies for the inconvenience, we've just changed the repo structure dramatically, and it may have some rough edges

pirj:
💚

benoittgt (commented):
LGTM. Thanks !

rspec-core/lib/rspec/core/runner.rb
benoittgt: Good catch

JonRowe:
Apologies for the delay in getting back to this some build failures and work have sucked up all my free time, its on my list to get this in this week.

rspec-core/lib/rspec/core/configuration.rb
JonRowe: Last change, can we wrap this up into the above

JonRowe:
Can we also back out the unrelated changes for the other specs, @pirj if you'd like to work on another PR for those I'm happy to review that.

rspec-core/lib/rspec/core/configuration.rb
byroot: I'm sorry, I don't think I understand what you mean by that.

pirj:
@JonRowe Done

pirj:
RuboCop job fix https://github.com/rspec/rspec/pull/201

HunkGenerator is [flaky](https://github.com/rspec/rspec-support/pull/599#issuecomment-2191807931) 

@JonRowe is the rest good to merge?

JonRowe:
Sorry for the silence on this, I'm going to merge this manually when I get the time to split it into 3.x and 4.x capable patches, but no input is needed beyond that.

rspec-core/lib/rspec/core/configuration.rb
pirj: @JonRowe I've tried inlining the `deprecation_warnings=` into `warnings=`, it ended up like this:

```diff
- self.deprecation_warnings = true
+ ::Warning[:deprecated] = true if defined?(::Warning) && ::Warning.respond_to?(:[]=)
```

Do you want to keep it as it is now, or inline?

The decision point here would be the ability to notice and remove those redundant checks when we drop Ruby below [2.7](https://ruby-doc.org/core-2.7.0/Warning.html#method-c-5B-5D-3D), correct?

I can add a commit here real quick.

pirj:
@JonRowe CI is green.
I'll make sure to replicate this to 4.0 once done with [CI for 4.0](https://github.com/rspec/rspec/pull/245).

rspec-core/lib/rspec/core/configuration.rb
```diff
@@ -1765,14 +1766,57 @@ def self.delegate_to_ordering_manager(*methods)
 
       # Set Ruby warnings on or off.
```
pirj: Note to self: this is outdated, and we need to document that we accept :none, :all or :deprecations_only.

rspec-core/lib/rspec/core/configuration.rb
```diff
@@ -1765,14 +1766,57 @@ def self.delegate_to_ordering_manager(*methods)
...
       def warnings?
         $VERBOSE
       end
 
+      # @private
+      def warnings_set?
+        @warnings_set
+      end
+
+      if defined?(::Warning) && ::Warning.respond_to?(:[]=)
```
pirj: Note to self: This is outdated, as we're dropping Ruby 2.x support, and can be simplified.

rspec-core/spec/rspec/core/runner_spec.rb
```diff
@@ -363,6 +363,19 @@ def interrupt
         expect(runner.instance_exec { @options }).to be(config_options)
       end
 
+      it "emits a warning if config.deprecation_warnings isn't set" do
```
pirj:
```suggestion
      it "emits a warning if config.warnings isn't set" do
```

rspec-core/lib/rspec/core/project_initializer/spec/spec_helper.rb
```diff
@@ -44,6 +44,11 @@
   # triggering implicit auto-inclusion in groups with matching metadata.
   config.shared_context_metadata_behavior = :apply_to_host_groups
 
+  # This setting enables Ruby deprecation warnings. It's recommended, but in some cases may
+  # be too noisy due to issues in dependencies. It can be set to `:all` or `:none`
+  # to either display all warnings or none of them.
+  config.warnings = :deprecations_only
```
pirj: Previously, we had `config.warnings = true`, which translates to `config.warnings = :all`.
Having `:deprecations_only` here is a setback.

[Here](https://github.com/rspec/rspec/issues/37#issuecomment-2504702317) we agree that turning all by default:
> place it into our generated spec helper, with a comment explaining what it is. This has the effect of effectively being on by default
```suggestion
  config.warnings = :all
```

rspec-core/lib/rspec/core/configuration.rb
pirj: My previous comment doesn't stand with us retiring Ruby 2.x support.
There are no downsides left from collapsing those two methods into one.

pirj:
I stumbled across our CLI `--warnings` option https://github.com/rspec/rspec/blob/029e58c4bdd75fe42f1034014145c4167a229f29/rspec-core/features/command_line/warnings_option.feature#L3

`--warnings` already works like `:all`.

rspec-core/lib/rspec/core/project_initializer/spec/spec_helper.rb
pirj: @JonRowe How do you feel about defaulting to `:all` in the project initializer?

pirj:
The code to warn users to set the configuration option is getting complicated with the text config (`.rspec`) and CLI options sparser involved.
So far, I made it work with such a check:
```ruby
      def warnings_set?
        @warnings_set || @preferred_options.key?(:warnings)
```
The second part is a shortcut, where a longer path would be to create some WarningsConfigurationManager, following the FilterManager and Order::ConfigurationManager, which feels overkill for what it does.

`warnings_option.feature` still fails. It expects no warnings if it doesn't pass `--warning` to `rspec`, but there's this warning to call `config.warning=` from `spec_helper.rb`. If we'd follow the path of adding this setup to this feature spec, whatever we set there undermines the elegance and the clarity of the spec.

What do you think of not warning when this option is not set, and default to `:all`?

I've pushed a temporary commit, but I'm not happy with the result, and would prefer taking another approach.

rspec-core/spec/spec_helper.rb
```diff
@@ -8,6 +8,7 @@ def self.new(*args, &block)
       # We detect ruby warnings via $stderr,
       # so direct our deprecations to $stdout instead.
       config.deprecation_stream = $stdout
+      config.warnings = :all
```
pirj: This is unnecessary, as [there's](https://github.com/rspec/rspec/blob/e4a549d062d20ad5ef5c98c0b28fdc613d200287/rspec-core/.rspec#L2) `--warnings` in `.rspec`.

It doesn't hurt to have both, but duplicating the logic that sets `$VERBOSE`/`Warning[:deprecated]` seems unnecessary.

rspec-core/lib/rspec/core/option_parser.rb
```diff
@@ -178,11 +178,8 @@ def parser(options)
...
 
         parser.on('-w', '--warnings', 'Enable ruby warnings') do
-          if Object.const_defined?(:Warning) && Warning.respond_to?(:[]=)
-            # :nocov: on older Ruby without Warning
-            Warning[:deprecated] = true
-            # :nocov:
-          end
+          options[:warnings] = :all
+          Warning[:deprecated] = true
           $VERBOSE = true
```
pirj, line 182: I fought the temptation to remove this and solely rely on `options[:warnings] = :all`, but there's [an example](https://github.com/rspec/rspec/blob/e4a549d062d20ad5ef5c98c0b28fdc613d200287/rspec-core/spec/rspec/core/option_parser_spec.rb#L392) named "      immediately enables warnings so that **warnings are issued for files loaded by `--require`**". That sounds like this option would be able to turn on warnings on early loading stage, before the `Configuration` is set up and modifies `$VERBOSE`/`Warning[:deprecated]`.

If it is so, this code needs a better explanation.

JonRowe: This could be re-worked at a later date, but would need to do some weird delayed dance to be consistent.

These run here because its the earliest points this config can be processed, it emulates the historic ruby warning flag, and yes as the example says, means warnings are reported for requires.

rspec-core/lib/rspec/core/project_initializer/spec/spec_helper.rb
JonRowe: The intent for 4 was to have the default set to `:deprecations_only` as recommended, so I moved the initialiser to suggest `:all` as an alternative.

rspec-core/spec/spec_helper.rb
JonRowe: It doesn't have the same effect, this is for other created instances, however it causes chicken and egg testing issues, and actually the outer config is now reflected due to the warnings check. That can probably be improved in future.

JonRowe:
I've finally gotten 4.0 and 3.99 into a place where this can be merged so I've tidied it up and done so :joy: Thanks for this @byroot I'm sorry its taken so long to land but the default warning behaviour in RSpec 4 will now be the equivalent of `:deprecations_only` with a recommendation to go to `:all` in the commented out project template.

PR #163, opened 2024-12-22
Be ready for Ruby 3.4 🎄

benoittgt:
Preparing the release of Ruby 3.4.

We need to adapt a little bit the ouput of RSpec matchers to avoid having many changes on too many files:
Ruby 3.3
```
irb(main):003> [{ :a => [1, 2] }].to_s
=> "[{:a=>[1, 2]}]"
```

Ruby 3.4
```
irb(main):001> [{ :a => [1, 2] }].to_s
=> "[{a: [1, 2]}]"
```

I find this PR quite dirty. I suggested something in https://github.com/rspec/rspec/pull/163#discussion_r1895077434

Too many `if` in one PR.

Simplecov doesn't seem to be happy with RUBY_VERSION conditions. [Normally coverage are merged](https://github.com/simplecov-ruby/simplecov/issues/634) should be merged at the end. So we have to modify the CI script.


.github/workflows/ci.yml
```diff
@@ -14,31 +14,31 @@ permissions:
 
 jobs:
   docs:
-    uses: rspec/rspec/.github/workflows/documentation_coverage.yml@main
+    uses: benoittgt/rspec-1/.github/workflows/documentation_coverage.yml@ruby-3-4
```
benoittgt, line 17: To be reverted. :) 

I had no push right on rspec/rspec so I used a fork. :)

rspec-expectations/lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -137,8 +137,18 @@ module HashFormatting
           #     {:a => 5, :b => 2}
           #
           # This is idempotent and safe to run on a string multiple times.
+          # On Ruby 3.4 and later, it also adds spaces around the hash rocket
+          # and converts `a: 1` to `:a => 1` for consistency.
           def improve_hash_formatting(inspect_string)
-            inspect_string.gsub(/(\S)=>(\S)/, '\1 => \2')
+            if RUBY_VERSION.to_f >= 3.4
+              if inspect_string.match?(/(\w+):/)
```
benoittgt: This is for me quite ugly. :/  I am not very confident with this change.

We have to do this in order to stay as much as possible compatible with previous RSpec output. But the thing is we are going the opposite way of the new hash output in Ruby 3.4.

I am wondering if we should be more flexible on the spec output in our reporting and simply have helper to match in our own specs the 3 formats:
- `{:a=>1}`
- `{:a => 1}`
- `{a: 1}`

Maybe something to change in dedicated version ?

rspec-expectations/lib/rspec/matchers/built_in/start_or_end_with.rb
```diff
@@ -29,7 +29,7 @@ def failure_message
         def description
           return super unless Hash === expected
           english_name = EnglishPhrasing.split_words(self.class.matcher_name)
-          description_of_expected = surface_descriptions_in(expected).inspect
+          description_of_expected = HashFormatting.improve_hash_formatting(surface_descriptions_in(expected).inspect)
```
benoittgt:
```
`hash_inspect` is not available on an example group
```

rspec-expectations/lib/rspec/matchers/built_in/start_or_end_with.rb
```diff
@@ -29,7 +29,11 @@ def failure_message
         def description
           return super unless Hash === expected
           english_name = EnglishPhrasing.split_words(self.class.matcher_name)
-          description_of_expected = surface_descriptions_in(expected).inspect
+          if RUBY_VERSION.to_f >= 3.4
+            description_of_expected = HashFormatting.improve_hash_formatting(surface_descriptions_in(expected).inspect)
```
benoittgt, line 33: I'm curious if we want to use this or a simpler `gsub`.

.github/workflows/ci.yml
JonRowe: I will look into that.

JonRowe:
Closing in favour of #164 thanks for kicking this off!

PR #164, opened 2024-12-26
Fix hash output on Ruby 3.4

JonRowe:
This builds off #163 but contains a fix which means Ruby 3.4 syntax will consistently use `{a: :b}` style on single lines.


benoittgt:
Love this proposal. :)

PR #165, opened 2024-12-28
Cancel outdated builds to reduce queueing

pirj:
Run CI only for the last push on each branch.

Prevents branch builds from stacking on multiple subsequent pushes by canceling older builds.

https://docs.github.com/en/actions/using-jobs/using-concurrency
https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/

Scavenged from https://github.com/rubocop/rubocop-rspec/pull/1539

Problem:
image

### DIFF
--- a/.github/workflows/documentation_coverage.yml
+++ b/.github/workflows/documentation_coverage.yml
@@ -4,6 +4,10 @@ on: workflow_call
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 env:
   RSPEC_CI: true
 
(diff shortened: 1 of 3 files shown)


JonRowe:
Reverting because it doesn't work, apologies, should have insisted on checking it before merging but it causes only one of the gems to be run.

pirj:
It must be the “group” signature that messes things up, I’ll experiment with it. Thanks for reverting.

JonRowe:
I think each gem / rubocop needs to have a different name (remembering that the rspec one runs multiple times) it looks like the workflow part of the name also is blank hence also caught rubocop

pirj:
That makes sense since we always schedule the same set of builds jobs.

PR #172, opened 2025-01-15
Reuse the same ExamplesNotification to improve memoization

MaxLap:
I noticed that the ExamplesNotification instantiated 3 times in the `finish`. Considering there is memoization going on in it for the notifications related methods, it seems logical that the same instance should be used all 3 times.

I removed the `stop` method which appears to be used only once, is quite trivial and has its own instantiation.

### DIFF
--- a/rspec-core/lib/rspec/core/reporter.rb
+++ b/rspec-core/lib/rspec/core/reporter.rb
@@ -172,10 +172,11 @@ module RSpec::Core
     # @private
     def finish
       close_after do
-        stop
+        examples_notification = Notifications::ExamplesNotification.new(self)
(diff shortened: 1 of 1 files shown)


MaxLap:
(I don't think the CI errors are related to my change)

JonRowe:
Can you rebase this, main should be fixed. I'm on the fence about dropping the method, you could always pass in the notification in the mean time.

MaxLap:
Rebased. That error is probably not from my change either

JonRowe:
Thanks for your patience here!

JonRowe:
Released in 3.13.3

PR #173, opened 2025-01-16
Fix memoization in #location_rerun_argument

MaxLap:
The use of `return` meant that the memoization just didn't happen as control flow left the whole `#location_rerun_argument` method. Using break instead fixes that.


MaxLap:
(I don't think the CI errors are related to my change)

JonRowe:
Can you rebase this, main should be fixed.

MaxLap:
Rebased

JonRowe:
Thanks, sorry for the delay I wanted to verify this as there wasn't a spec.

JonRowe:
Released in 3.13.3

PR #177, opened 2025-01-29
[rspec-expectations] update not to invalid matcher error to be more descriptive

### LINKED ISSUE #90: "NotImplementedError: `expect { }.not_to change { }.by()` is not supported" - would be nice to improve
Hi, there is an error when you add `.by(..)` to your `expect {}.not_to change(some, :attr)`, like `expect {}.not_to change(some, :attr).by(1)` that says "NotImplementedError: `expect { }.not_to change { }.by()` is not supported" and.., it's really not good, because you see the first things that say "You can use `change` matcher with the block" but you miss that there is a problem 'cause you forget to delete or just added `.by(..)` to your matcher.
My solution:
To change the error message like to this: `Hey, you cannot use '.by(1)' in your <code> matcher, it is not supported, delete it or change matcher, please`.
*Would be fun if the message can be like this, please ^-^

QingyuChai:
Fixes #90

update invalid matcher error for 'expect { }.not_to change { }.<matcher>' to be more descriptive

Error thrown:
| Before | After|
| ----- | -----|
| image | image|

Very much a ruby noob, and ran into a similar issue as here:
https://github.com/rspec/rspec/issues/90

### DIFF
--- a/rspec-expectations/lib/rspec/matchers/built_in/change.rb
+++ b/rspec-expectations/lib/rspec/matchers/built_in/change.rb
@@ -286,7 +286,7 @@ module RSpec
         # @private
         def does_not_match?(event_proc)
           if @description_suffix
-            raise NotImplementedError, "Using `expect { }.not_to change { }` " \
+            raise NotImplementedError, "Using `expect { }.not_to change { }.from()` " \
                                        "with the `to()` matcher is not supported."
           end
 
--- a/rspec-expectations/lib/rspec/matchers/built_in/change.rb
+++ b/rspec-expectations/lib/rspec/matchers/built_in/change.rb
@@ -150,8 +150,8 @@ module RSpec
 
         # @private
         def does_not_match?(_event_proc)
-          raise NotImplementedError, "`expect { }.not_to change " \
-                                     "{ }.#{@relativity}()` is not supported"
+          raise NotImplementedError, "Using `expect { }.not_to change { }` " \
+                                     "with the `#{@relativity}()` matcher is not supported."
         end
 
         # @private
@@ -286,8 +286,8 @@ module RSpec
         # @private
         def does_not_match?(event_proc)
           if @description_suffix
-            raise NotImplementedError, "`expect { }.not_to change { }.to()` " \
-                                       "is not supported"
+            raise NotImplementedError, "Using `expect { }.not_to change { } ` " \
+                                       "with the `to()` matcher is not supported."
           end
 
           perform_change(event_proc) && !@change_details.changed? && @matches_before
@@ -326,8 +326,8 @@ module RSpec
 
         # @private
         def does_not_match?(_event_proc)
-          raise NotImplementedError, "`expect { }.not_to change { }.to()` " \
-                                     "is not supported"
+          raise NotImplementedError, "Using `expect { }.not_to change { }` " \
+                                     "with the `to()` matcher is not supported."
         end
 
       private
--- a/rspec-expectations/lib/rspec/matchers/built_in/change.rb
+++ b/rspec-expectations/lib/rspec/matchers/built_in/change.rb
@@ -152,7 +152,6 @@ module RSpec
         def does_not_match?(_event_proc)
           raise NotImplementedError, "Using a negated form of the `change` matcher " \
                                      "with `#{@relativity}()` is not supported."
-
         end
 
         # @private
--- a/rspec-expectations/lib/rspec/matchers/built_in/change.rb
+++ b/rspec-expectations/lib/rspec/matchers/built_in/change.rb
@@ -150,8 +150,9 @@ module RSpec
 
         # @private
         def does_not_match?(_event_proc)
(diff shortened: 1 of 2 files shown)


rspec-expectations/lib/rspec/matchers/built_in/change.rb
```diff
@@ -286,8 +286,8 @@ def to(value)
         # @private
         def does_not_match?(event_proc)
           if @description_suffix
-            raise NotImplementedError, "`expect { }.not_to change { }.to()` " \
-                                       "is not supported"
+            raise NotImplementedError, "Using `expect { }.not_to change { }` " \
+                                       "with the `to()` matcher is not supported."
```
pirj: This should actually say `from()`, shouldn't it? Apparently this was a copy-pasta.

QingyuChai: ah, nice catch, I dug a little into this and found out that it should indeed be `to()`, but the `to()` has to be after a `from()`

e.g.

Supported:
`expect {}.not_to change {}.from(5)`

Not supported:
`expect {}.not_to change {}.from(5).to(6)`

New error message:
>  Using `expect { }.not_to change { }.from()` with the `to()` matcher is not supported.

QingyuChai: I'm open to suggestions for the error message as well - I came up with it on the fly

pirj: What you say makes sense. There's nothing wrong with `expect { }.not_to change { }.from()`, we're just explicitly setting an expectation for the initial value, and expect it not to change.

It turns out that the code responsible for `change { }.from().to()` is different from the one responsible for `change { }.to().from()`, and I can't quickly grasp why this is done like this.

But this whole thing is orthogonal to your change, so please never mind the above.

pirj: See #178

rspec-expectations/lib/rspec/matchers/built_in/change.rb
```diff
@@ -150,8 +150,8 @@ def matches?(event_proc)
 
         # @private
         def does_not_match?(_event_proc)
-          raise NotImplementedError, "`expect { }.not_to change " \
-                                     "{ }.#{@relativity}()` is not supported"
+          raise NotImplementedError, "Using `expect { }.not_to change { }` " \
+                                     "with the `#{@relativity}()` matcher is not supported."
```
pirj: Strictly speaking, `by()` is not a matcher. I've seen those were mentioned as "qualifiers". The `change` is the matcher here.


WDYT of "Using a negated form of the `change` matcher with a `to()` qualifier is not supported."?

pirj: One other case is when the negation happens not with a `not_to`/`to_not`, but rather by creating a negated version with `define_negated_matcher :not_change, :change`.

The suggested error message would be less confusing for a negated matcher.

rspec-expectations/lib/rspec/matchers/built_in/change.rb
```diff
@@ -326,8 +326,8 @@ def from(value)
 
         # @private
         def does_not_match?(_event_proc)
-          raise NotImplementedError, "`expect { }.not_to change { }.to()` " \
-                                     "is not supported"
+          raise NotImplementedError, "Using `expect { }.not_to change { }` " \
```
pirj: We've changed the failure messages, but no specs failed.
May I kindly ask you to add such specs for the `.to()` and `.by()` cases?

rspec-expectations/lib/rspec/matchers/built_in/change.rb
QingyuChai: That sounds good to me, it's clearly pointing out the error so I'm happy with that

rspec-expectations/lib/rspec/matchers/built_in/change.rb
QingyuChai: Yeah, I can do that. I'll wait for this to merge first?
https://github.com/rspec/rspec/pull/178

pirj: It’s fine, I’ll adjust and rebase that other pr after yours. With specs for error messages this will be an easy thing to do.

rspec-expectations/lib/rspec/matchers/built_in/change.rb
pirj: Even “ Using a negated form of the `change` matcher with `to()` is not supported.” is good, without all the too-technical terms involved.

QingyuChai: yeah I like that, no need for complicated lingo!

pirj (approved):
Perfect, thank you! 🤩

QingyuChai:
Thank YOU! Appreciate the quick responses

JonRowe (approved):

PR #178, opened 2025-01-29
Join the change {}.to().from() matchers

pirj:
Previously, those two were implemented separately

Follow-up to #177

### DIFF
--- a/rspec-expectations/lib/rspec/matchers/built_in/change.rb
+++ b/rspec-expectations/lib/rspec/matchers/built_in/change.rb
@@ -34,13 +34,13 @@ module RSpec
         # @api public
         # Specifies the new value you expect.
         def to(value)
-          ChangeToValue.new(change_details, value)
+          ChangeFromTo.new(change_details, ChangeFromTo::MATCH_ANYTHING, value)
         end
 
         # @api public
         # Specifies the original value.
         def from(value)
-          ChangeFromValue.new(change_details, value)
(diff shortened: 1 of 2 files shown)


rspec-expectations/lib/rspec/matchers/built_in/change.rb
```diff
@@ -190,16 +190,50 @@ def initialize(change_details, from, to)
           @expected_after  = to
         end
 
+        def to(to)
+          raise "Can't set `to()` more than once" unless @expected_after == MATCH_ANYTHING
```
pirj, line 196: This is new. And we'll need a spec for this.

JonRowe:
I don't feel like this improves the code, it adds a bunch of conditionals that are better structured as the two separate classes, is there any reason why you feel this is better?

pirj:
Those conditionals were previously implicitly present as “…#{description_suffix}”. To me, this was cryptic.

The main problem here is that it looked like those qualifiers can be chained arbitrarily, like `from(3).by(5)`, but they can’t, because their implementation is very specific. This is the only problem ai’ve tackled here, removing common parts and ordering the from/to parts of the description.

From my point of view, the code now reflects the public interface more clearly, and there’s less duplication.

As you mention, those qualifiers could be better all implemented separately. But this is a more involving rework which I would prefer to leave for later.

JonRowe:
I would prefer this remain as two classes, if you'd like to improve the chaining issue in that format feel free but I don't want refactor this into conditionals over the explicitness of the two classes.

PR #183, opened 2025-02-13
[rspec-expectations] Fix message when matching array with non-array

jrmhaig:
#### Description

Update the failure message of the `match` matcher when an array is expected and the actual value is not an array.

#### Why

Currently

```ruby
  expect('string').to match([1,2,3])
```

results in an exception at `actual.match(expected)` in `RSpec::Matchers::BuiltIn::Match#match`. This is not caught correctly by `can_safely_call_match?`.

#### How

`RSpec::Matchers::BuiltIn::Match#match` returns `false` when `values_match?` fails and `expected` is an array. A `failure_message` method is added to catch this case and return a suitable message.

I am tipping my toe in the water with something simple to try to understand the internals of RSpec. I had originally started looking at #143 (which I will continue to look at) and found this in the process, which appears to be a bug.

### DIFF
--- a/rspec-expectations/lib/rspec/matchers/built_in/match.rb
+++ b/rspec-expectations/lib/rspec/matchers/built_in/match.rb
@@ -35,11 +35,23 @@ module RSpec
           self
         end
 
+        # @api private
+        # @return [String]
+        def failure_message
+          if Array === expected && !(actual.respond_to?(:to_a) || actual.respond_to?(:to_ary))
+            return "expected a collection that can be converted to an array with " \
+                   "`#to_ary` or `#to_a`, but got #{actual_formatted}"
+          end
+
+          super
+        end
+
       private
 
         def match(expected, actual)
           return match_captures(expected, actual) if @expected_captures
           return true if values_match?(expected, actual)
+          return false if Array === expected
           return false unless can_safely_call_match?(expected, actual)
           actual.match(expected)
         end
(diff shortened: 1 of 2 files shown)


jrmhaig:
There are two tests failing in `rspec-mocks/spec/rspec/mocks/diffing_spec.rb`:

* spec/rspec/mocks/diffing_spec.rb:46
* spec/rspec/mocks/diffing_spec.rb:57

If I change the first one to see the message that is being checked:

```ruby
    it "prints a diff of the strings for individual mismatched multi-line string arguments" do
      with_unfulfilled_double do |d|
        expect(d).to receive(:foo).with("some string\nline2")
        d.foo("this other string")
      end
    end
```

I get the output;

```
       #<Double "double"> received :foo with unexpected arguments
         expected: ("some string\nline2")
              got: ("this other string")
       Diff:
       @@ -1,2 +1 @@
       -some string
       -line2
       +this other string
```

and the expected output is

```
       #<Double "double"> received :foo with unexpected arguments
         expected: ("some string\nline2")
              got: ("this other string")
       Diff:
       @@ -1,3 +1,2 @@
       -some string
       -line2
       +this other string
```

I see this on the main branch as well as this branch and the difference is in ` @@ -1,3 +1,2 @@`, which I don't fully understand. Can you clarify?

JonRowe:
:wave: It seems diff-lcs updated and "fixed" some diffs

jrmhaig:
> 👋 It seems diff-lcs updated and "fixed" some diffs

Thanks. I had dug down as far as seeing that the output was coming from `diff-lcs` but hadn't considered downgrading the version. I was about to pull another PR that I can see passing to check. This would almost certainly have also failed for me, thus confusing me even more. I've pinned the version for the moment so that the pipeline can get over that failure.

JonRowe:
I'm hoping #185 will fix the build and then this can be rebased

JonRowe:
#185 is merged, can you rebase and drop the gem pinning commit

JonRowe:
Released as 3.13.4

PR #184, opened 2025-02-15
Fix tests for newer version of diff-lcs

jrmhaig:
Version 1.6 of the diff-lcs gem slightly alters the output resulting in two tests failing. The tests have been updated and the version of diff-lcs is pinned to '-> 1.6'.


JonRowe:
Thanks for looking at this, but in #185 I'm fixing the diffs because it will help me with 3.13 / 3.99 release even though 4.0 will pin to 1.6 and fix the diffs that way.

PR #187, opened 2025-02-15
Adds diff-lcs back to the matrix for 3-13-maintenance

JonRowe:


JonRowe:
This was cherry-picked to 3-13-maintenance

PR #189, opened 2025-02-16
[rspec-expectations] Add more detailed array diff details to `match` failure message

### LINKED ISSUE #143: `match` diff output should look like `match_array` when matching arrays
### Subject of the issue
`expect(["a", "b", "c"]).to match(["a", "b"])` should print unexpected extra elements in a nice way like `expect(["a", "b", "c"]).to match_array(["a", "b"])`

Since `match` checks order while `match_array` does not, there is no way to get pretty diffs _and_ check order of an array and get a nice diff. For very large matches (like API response matching) this makes the failure message hard to use. [See this gist for an example](https://gist.github.com/alan-pie/89eb98882c2de781c03ba2800a5fd2b5)

### Your environment
* Ruby version: 3.1.4
* rspec-expectations version: 3.12.2

### Steps to reproduce
```
RSpec.describe 'matching arrays' do
  it 'returns a nice diff when there are unexpected extra elements' do
    expect(["a", "b"]).to match(["a"])
  end
end
```

### Expected behavior
```
expected collection contained:  ["a"]
actual collection contained:    ["a", "b"]
the extra elements were:        ["b"]
```

### Actual behavior
`expected ["a", "b", "c"] to match ["a", "b"]`

jrmhaig:
#### What

Update the `match` built-in matcher to give more details of differences with arrays.

#### Why

Fixes #143 

`contain_exactly` and `match_array` both show diffs between arrays as:

```
expected collection contained:  ["a"]
actual collection contained:    ["a", "b"]
the extra elements were:        ["b"]
```

and this makes it clear what needs to be fixed. the `array` matcher, which primarily works with strings and regular expressions, instead returns:

```
expected ["a", "b", "c"] to match ["a", "b"]
```

For consistency the output is updated to match that given by `contain_exactly` and `match_array`.

#### How

`rspec-expectations/lib/rspec/matchers/built_in/contain_exactly.rb` includes the logic to calculate the differences between arrays, including expected arrays. This is extracted into a helper class, `rspec-expectations/lib/rspec/matchers/built_in/helpers/array_matchers.rb`, that is then used by both `contain_exactly.rb` and `match.rb`.

The arrays in the output for the `match` matcher are not sorted as the ordering is important. For the same reason, the wording is altered for `match.rb` slightly to `expected collection was` and `actual collection was`.


jrmhaig:
Hi @JonRowe. Is there anything that needs to be done to this? This is my first significant PR so I appreciate that I may not have done things correctly regarding the design decisions of the project.

jrmhaig:
Hi @JonRowe. Do you have any comment on this? Is it of interest or should I close it? I see that there is now a conflict to be resolved and I think I remember enough about this to try to resolve it.

PR #192, opened 2025-02-27
[rspec-core] Ensure $SHELL is correct before tests

### LINKED ISSUE #190: [rspec-core] Tests in spec/integration/location_rerun_spec.rb pass or fail depending on shell
When I execute the tests in `rspec-core/spec/integration/location_rerun_spec.rb` I get failures:

```
rspec ./spec/integration/location_rerun_spec.rb:66 # Failed spec rerun location when config.force_line_number_for_spec_rerun is set to false prints the example id of the failed assertion
rspec ./spec/integration/location_rerun_spec.rb:80 # Failed spec rerun location when config.force_line_number_for_spec_rerun is set to false and the shared examples are defined in a separate file prints the example id of the failed assertion
```

I am fairly certain that this is because the formatting of the `rspec` command in the output depends on the shell. The test expects:

```
rspec './local_shared_examples_spec.rb[1:1:1:1]' #  the first context behaves like a failing spec fails
```

and I get

```
rspec ./local_shared_examples_spec.rb[1:1:1:1] #  the first context behaves like a failing spec fails
```

as the shell I am using (bash on Ubuntu) does not require quotes around a string with square brackets.

I don't know if this is a problem with the test or a bug with the generation of the output. I think that always including quote marks in the output is safest as this will still work even when they are not required. I don't know how easy it would be to implement this, though.

**Update**

I have just found this

https://github.com/rspec/rspec/blob/6decbce170ab0bec88f461ae3bd75f65115fc54d/rspec-core/lib/rspec/core/shell_escape.rb#L32-L46

which suggests that this is working as intended so perhaps the test needs to stub the value of `ENV['SHELL']`.

jrmhaig:
Fixes #190 

The formatting of the example rerun command depends on $SHELL so these tests can fail depending on the user's settings unless the variable is stubbed.

### DIFF
--- a/rspec-core/spec/integration/location_rerun_spec.rb
+++ b/rspec-core/spec/integration/location_rerun_spec.rb
@@ -63,6 +63,8 @@ RSpec.describe 'Failed spec rerun location' do
   end
 
   context "when config.force_line_number_for_spec_rerun is set to false" do
+    around { |ex| with_env_vars('SHELL' => '/usr/local/bin/zsh', &ex) }
+ 
     it 'prints the example id of the failed assertion' do
       run_command("#{Dir.pwd}/tmp/aruba/local_shared_examples_spec.rb")
 
--- a/rspec-core/spec/integration/location_rerun_spec.rb
+++ b/rspec-core/spec/integration/location_rerun_spec.rb
@@ -64,7 +64,7 @@ RSpec.describe 'Failed spec rerun location' do
 
   context "when config.force_line_number_for_spec_rerun is set to false" do
     around { |ex| with_env_vars('SHELL' => '/usr/local/bin/zsh', &ex) }
- 
+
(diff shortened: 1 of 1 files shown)


rspec-core/spec/integration/location_rerun_spec.rb
```diff
@@ -63,6 +63,8 @@
   end
 
   context "when config.force_line_number_for_spec_rerun is set to false" do
+    around { |ex| with_env_vars('SHELL' => '/usr/local/bin/zsh', &ex) }
+
```
JonRowe:
```suggestion
    around { |ex| with_env_vars('SHELL' => '/usr/local/bin/zsh', &ex) }

```

JonRowe:
Thanks, I've merged this despite the rubocop failure as that is unrelated

PR #193, opened 2025-02-28
Show the exception detailed message when available

### LINKED ISSUE #191: Show detailed message when showing an exception
Since Ruby 3.2's `did_you_mean` and `error_highlight` gems stopped modifying `Exception#message`, the RSpec output stopped showing some valuable feedback that is now available in the `Exception#detailed_message` method.

For example, for the following class and spec:

```rb
require "spec_helper"

class Cat
  def yell
    "meow".upcsae # Notice the typo
  end
end

RSpec.describe Cat do
  describe "#yell" do
    it "returns a loud meow" do
      cat = Cat.new
      expect(cat.yell).to eq("MEOW")
    end
  end
end
```

Ruby 3.1.6 output is:

```
  1) Cat#yell returns a loud meow
     Failure/Error: "meow".upcsae

     NoMethodError:
       undefined method `upcsae' for "meow":String

           "meow".upcsae
                 ^^^^^^^
       Did you mean?  upcase
     # ./spec/models/cat_spec.rb:5:in `yell'
     # ./spec/models/cat_spec.rb:13:in `block (3 levels) in <top (required)>'

Failed examples:

rspec ./spec/models/cat_spec.rb:11 # Cat#yell returns a loud meow
```

Ruby 3.4.2 (actually, any Ruby version > 3.2.0), on the other hand, outputs:
```
  1) Cat#yell returns a loud meow
     Failure/Error: "meow".upcsae

     NoMethodError:
       undefined method 'upcsae' for an instance of String
     # ./spec/models/cat_spec.rb:5:in 'Cat#yell'
     # ./spec/models/cat_spec.rb:13:in 'block (3 levels) in <top (required)>'
```

RSpec could detect if `Exception#detailed_message` is defined and show it instead of `Exception#message`.

neilvcarvalho:
Ruby 3.2 introduced an `Exception#detailed_message` method to allow gems such as `did_you_mean` and `error_highlight` to enrich the exception message.  `Exception#detailed_message`, by default, shows the error message, the exception class name, and optional highlighting using ANSI escape sequences. Previously, those gems modified `Exception#message`, which led to broken test suites.

The problem with this change is that, as we only show `Exception#message`, we stopped showing valuable information, such as typo corrections and error locations.

This commit reintroduces this feedback by using the `Exception#detailed_message` whenever available.

After this change, exceptions with detailed messages will be shown as:

```
     NoMethodError:
       undefined method `upcsae' for an instance of String

           "meow".upcsae
                 ^^^^^^^
       Did you mean?  upcase
```

Closes #191

### DIFF
--- a/rspec-core/lib/rspec/core/formatters/exception_presenter.rb
+++ b/rspec-core/lib/rspec/core/formatters/exception_presenter.rb
@@ -184,7 +184,13 @@ module RSpec
 
         # rubocop:disable Lint/RescueException
         # :nocov:
-        if SyntaxError.instance_methods.include?(:detailed_message)
+        if RSpec::Support::RubyFeatures.supports_exception_detailed_message?
+          def exception_message_string(exception)
+            exception.detailed_message(:highlight => false).sub(" (#{exception.class})", '')
+          rescue Exception => other
+            "A #{exception.class} for which `exception.detailed_message.to_s` raises #{other.class}."
+          end
+        elsif SyntaxError.instance_methods.include?(:detailed_message)
           def exception_message_string(exception)
             case exception
             when SyntaxError then exception.detailed_message.to_s
@@ -285,7 +291,7 @@ module RSpec
           end
         end
 
-        if  String.method_defined?(:encoding)
+        if String.method_defined?(:encoding)
           def encoded_description(description)
             return if description.nil?
             encoded_string(description)
--- a/rspec-core/spec/rspec/core/formatters/exception_presenter_spec.rb
+++ b/rspec-core/spec/rspec/core/formatters/exception_presenter_spec.rb
@@ -340,11 +340,17 @@ module RSpec::Core
 
         the_presenter = Formatters::ExceptionPresenter.new(incorrect_message_exception, example)
 
+        statement = if RSpec::Support::RubyFeatures.supports_exception_detailed_message?
+                      'exception.detailed_message.to_s'
+                    else
+                      'exception.message.to_s'
+                    end
+
         expect(the_presenter.fully_formatted(1)).to eq(<<-EOS.gsub(/^ +\|/, ''))
           |
           |  1) Example
           |     Failure/Error: Unable to find matching line from backtrace
-          |       A #{FakeException} for which `exception.message.to_s` raises #{expected_error}.
+          |       A #{FakeException} for which `#{statement}` raises #{expected_error}.
         EOS
       end
 
@@ -568,6 +574,27 @@ module RSpec::Core
               |     # ./spec/rspec/core/formatters/exception_presenter_spec.rb:
             EOS
           end
+        end
+      end
+
+      context "with detailed_message support", :if => RSpec::Support::RubyFeatures.supports_exception_detailed_message? do
+        class ExceptionWithDetailedMessage < StandardError
+          def detailed_message(_options = {})
+            "Detailed error message (#{self.class})"
+          end
+
+          def message
+            "Regular message"
+          end
+        end
+
+        let(:detailed_exception) { ExceptionWithDetailedMessage.new }
+        let(:detailed_presenter) { RSpec::Core::Formatters::ExceptionPresenter.new(detailed_exception, example) }
+
+        it "uses detailed_message when available" do
+          formatted_output = detailed_presenter.fully_formatted(1)
+          expect(formatted_output).to include("Detailed error message")
+          expect(formatted_output).not_to include("Regular message")
         end
       end
     end
--- a/rspec-core/spec/rspec/core/notifications_spec.rb
+++ b/rspec-core/spec/rspec/core/notifications_spec.rb
@@ -12,7 +12,11 @@ RSpec.describe "FailedExampleNotification" do
 
   let(:example) { new_example(:status => :failed) }
   exception_line = __LINE__ + 1
-  let(:exception) { instance_double(Exception, :backtrace => [ "#{__FILE__}:#{exception_line}"], :message => 'Test exception') }
+  if RSpec::Support::RubyFeatures.supports_exception_detailed_message?
+    let(:exception) { instance_double(Exception, :backtrace => [ "#{__FILE__}:#{exception_line}"], :message => 'Test exception', :detailed_message => 'Test exception') }
+  else
+    let(:exception) { instance_double(Exception, :backtrace => [ "#{__FILE__}:#{exception_line}"], :message => 'Test exception') }
+  end
   let(:notification) { ::RSpec::Core::Notifications::ExampleNotification.for(example) }
 
   before do
@@ -277,7 +281,6 @@ RSpec.describe "FailedExampleNotification" do
         end
       end
     end
-
     context "when the exception is a MultipleExceptionError" do
       let(:sub_failure_1)  { StandardError.new("foo").tap { |e| e.set_backtrace([]) } }
       let(:sub_failure_2)  { StandardError.new("bar").tap { |e| e.set_backtrace([]) } }
@@ -326,8 +329,10 @@ RSpec.describe "FailedExampleNotification" do
       it "returns failures_lines with invalid bytes replace by '?'" do
         message_with_invalid_byte_sequence =
           "\xEF \255 \xAD I have bad bytes".dup.force_encoding(Encoding::UTF_8)
-        allow(exception).to receive(:message).
-          and_return(message_with_invalid_byte_sequence)
+        allow(exception).to receive(:message).and_return(message_with_invalid_byte_sequence)
+        if RSpec::Support::RubyFeatures.supports_exception_detailed_message?
+          allow(exception).to receive(:detailed_message).and_return(message_with_invalid_byte_sequence)
+        end
 
         lines = notification.message_lines
         expect(lines[0]).to match %r{\AFailure\/Error}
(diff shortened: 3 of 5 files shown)


rspec-core/lib/rspec/core/formatters/exception_presenter.rb
```diff
@@ -184,13 +184,10 @@ def failure_slash_error_lines
...
-        if SyntaxError.instance_methods.include?(:detailed_message)
+        if RSpec::Support::RubyFeatures.supports_exception_detailed_message?
           def exception_message_string(exception)
-            case exception
-            when SyntaxError then exception.detailed_message.to_s
-            else
-              exception.message.to_s
-            end
+            puts "exception.class: #{exception.class}"
+            exception.detailed_message(:highlight => false).sub(/\s\(#{Regexp.escape(exception.class.to_s)}\)$/, '')
```
neilvcarvalho: I was concerned about this hash being converted into keyword arguments, which Ruby 2.7 removed, but no warnings were raised.

neilvcarvalho:
I had added some specs, and deleted them when I realized there was no difference between the specs that didn't include the detailed messages and the ones I had added. Many specs fail if the class name isn't removed from the detailed message, for example.

I would see value in a spec that stubs the `detailed_message` method and ensure it is shown instead of the regular message, though.

neilvcarvalho:
I added a spec specific for the detailed message on 2728c9f055920029a9be087485d3731110fdbfa9. It deviates a bit from other specs, but I think it does the job of testing that the detailed message is present.

rspec-core/spec/rspec/core/notifications_spec.rb
```diff
@@ -12,7 +12,7 @@ def self.wrap(text, code_or_symbol)
 
   let(:example) { new_example(:status => :failed) }
   exception_line = __LINE__ + 1
-  let(:exception) { instance_double(Exception, :backtrace => [ "#{__FILE__}:#{exception_line}"], :message => 'Test exception') }
+  let(:exception) { instance_double(Exception, :backtrace => [ "#{__FILE__}:#{exception_line}"], :message => 'Test exception', :detailed_message => 'Test exception') }
```
pirj: Would this fix the failing specs?

```suggestion
  if RSpec::Support::RubyFeatures.supports_exception_detailed_message?
    let(:exception) { instance_double(Exception, :backtrace => [ "#{__FILE__}:#{exception_line}"], :message => 'Test exception', :detailed_message => 'Test exception') }
  else
    let(:exception) { instance_double(Exception, :backtrace => [ "#{__FILE__}:#{exception_line}"], :message => 'Test exception') }
  end  
```

rspec-core/spec/rspec/core/notifications_spec.rb
```diff
@@ -12,7 +12,11 @@ def self.wrap(text, code_or_symbol)
 
   let(:example) { new_example(:status => :failed) }
   exception_line = __LINE__ + 1
-  let(:exception) { instance_double(Exception, :backtrace => [ "#{__FILE__}:#{exception_line}"], :message => 'Test exception') }
+  if RSpec::Support::RubyFeatures.supports_exception_detailed_message?
+    let(:exception) { instance_double(Exception, :backtrace => [ "#{__FILE__}:#{exception_line}"], :message => 'Test exception', :detailed_message => 'Test exception') }
+  else
+    let(:exception) { instance_double(Exception, :backtrace => [ "#{__FILE__}:#{exception_line}"], :message => 'Test exception') }
+  end
```
pirj:
```suggestion
  end
```

pirj:
The last failure is tricky. If we stub the Exception's `detailed_message` to return invalid UTF8, it fails:
```
    A RSpec::Mocks::InstanceVerifyingDouble for which `exception.detailed_message.to_s` raises ArgumentError invalid byte sequence in UTF-8.
```

I suggest just muting this spec, as returning invalid UTF8 from `detailed_message` doesn't seem to be supported by Ruby.

neilvcarvalho:
I'm sorry I'm taking this long to work on this. I have some dedicated time
to open source work on Fridays, but I needed to use some to make up for
sick leaves or local holidays.

I tried reproducing the failures locally, but CI behaves differently
somehow. I could reproduce one in a Docker container, only for other
failure to happen in another Ruby version instead.

pirj:
It turns out to be caused by `.sub` that we call on that message with invalid UTF8.
With that, we're trying to retain the previous behaviour where we were not showing the exception class.
```diff
          1) Example
             Failure/Error: # The failure happened here! Handles encoding too! ЙЦ

       -       Boom
       +       Boom (RSpec::Core::FakeException)
               Bam
             # ./spec/rspec/core/formatters/exception_presenter_spec.rb:41
             # ------------------
             # --- Caused by: ---
       -     #   Real
       +     #   Real (RSpec::Core::FakeException)
             #   culprit
             #   ./spec/rspec/core/formatters/exception_presenter_spec.rb:184
```

I'd rather just follow what `detailed_message` provides, and avoid additional logic (`.sub`).

I also have some hiccups with running specs locally, e.g. without skipping spec/rspec/core/formatters/exception_presenter_spec.rb:329, `rspec` fails and doesn't even print the report properly.
This might be related to those changes, and needs to be looked at.

pirj:
On the other hand, 
```
        -                   expected pass, but baz
        +                   expected pass, but baz (RSpec::Expectations::ExpectationNotMetError)
```
only adds unnecessary noise.
It might make sense to remove the exception class in some cases, but not all.

The problem with the spec at line 329 was expecting `"exception.message.to_s"` in the message, while for new Rubies it should be `"exception.detailed_message.to_s"`.

rspec-core/spec/rspec/core/formatters/exception_presenter_spec.rb
```diff
@@ -340,11 +340,17 @@ def to_s
 
         the_presenter = Formatters::ExceptionPresenter.new(incorrect_message_exception, example)
 
+        statement =
```
neilvcarvalho: Oh, I like this. I was going to handle the failure message later, but forgot to do so. Your solution is elegant.

rspec-support/lib/rspec/support/ruby_features.rb
```diff
@@ -96,6 +96,16 @@ def supports_exception_cause?
...
 
+      if Exception.method_defined?(:detailed_message)
+        def supports_exception_detailed_message?
+          true
+        end
+      else
+        def supports_exception_detailed_message?
+          false
+        end
+      end
```
JonRowe, line 107: Can you break this out into a separate PR, the only reasoning I have for this is that it will allow me to merge it so we don't have to keep approving your workflows.

rspec-core/lib/rspec/core/formatters/exception_presenter.rb
```diff
@@ -184,15 +184,11 @@ def failure_slash_error_lines
 
         # rubocop:disable Lint/RescueException
         # :nocov:
-        if SyntaxError.instance_methods.include?(:detailed_message)
+        if RSpec::Support::RubyFeatures.supports_exception_detailed_message?
```
JonRowe, line 187: I would like to see this purely as an addition with a fallback to the syntax error check

rspec-core/lib/rspec/core/formatters/exception_presenter.rb
```diff
@@ -285,7 +281,7 @@ def formatted_message_and_backtrace(colorizer)
           end
         end
 
-        if  String.method_defined?(:encoding)
+        if String.method_defined?(:encoding)
```
JonRowe, line 294: Another alternative for opening a "small change" pr would be just this.

rspec-support/lib/rspec/support/ruby_features.rb
neilvcarvalho: Done on https://github.com/rspec/rspec/pull/203

rspec-core/lib/rspec/core/formatters/exception_presenter.rb
neilvcarvalho: Do you mean something like:

```rb
# New code
if RSpec::Support::RubyFeatures.supports_exception_detailed_message?
  def exception_message_string(exception)
    exception.detailed_message(:highlight => false)
  rescue Exception => other
    "A #{exception.class} for which `exception.detailed_message.to_s` raises #{other.class}."
  end
# Fallback to syntax error check
elsif SyntaxError.instance_methods.include?(:detailed_message)
  def exception_message_string(exception)
    case exception
    when SyntaxError then exception.detailed_message.to_s
    else
      exception.message.to_s
    end
  rescue Exception => other
    "A #{exception.class} for which `exception.message.to_s` raises #{other.class}."
  end
# Default case when theres no `detailed_message` at all
else
  def exception_message_string(exception)
    exception.message.to_s
  rescue Exception => other
    "A #{exception.class} for which `exception.message.to_s` raises #{other.class}."
  end
end
```

neilvcarvalho:
Phew. I think this is as ready as I can possibly make it. I tried hard to just deliver `detailed_message.to_s` without any substitution, but after fixing so many different specs with error message suffixes, I got stuck on the HTML formatter spec, which uses saved files.

neilvcarvalho:
@JonRowe @pirj Hey! This took me way longer than I anticipated, and I ended up going back to one of the initial implementations. I'd love to hear your feedback.

neilvcarvalho:
Hi @JonRowe @pirj! Is there anything I can do to move this PR forward?

PR #196, opened 2025-03-09
Remove duplicate `exist` listing from nav

dguardado:
[rspec.info](https://rspec.info/features/3-13/rspec-expectations/built-in-matchers/) seems to show two links for the `exist` matcher feature. This change removes the first listing and keeps the one that maintains alphabetical sort.

Seems to have been introduced with reordering in dfe834a so I assume this was unintentional.

Screenshot 2025-03-08 at 21-05-12


JonRowe:
Thanks!

PR #197, opened 2025-03-11
Include webrick gem for ruby > 3, necessary for yard docs

AnthonyClark:
When running `bundle exec yard server --reload` before:
```
.../3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require': cannot load such file -- webrick (LoadError)
```

Webrick is [no longer bundled as of Ruby 3](https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/#:~:text=sdbm-,webrick,-net%2Dtelnet), and yard needs it (or another server).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,8 @@ end
 group :documentation do
   gem 'github-markup', :platform => :mri
   gem 'redcarpet', :platform => :mri
+  # Webrick extracted from Ruby 3.0.0, required by Yard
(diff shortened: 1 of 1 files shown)


JonRowe:
Thanks, we tend to just build the docs and pump them over to the website but that flow isn't quite brought over for the monorepo, so theres no harm in adding this.

AnthonyClark:
@JonRowe makes sense - this was just something I noticed while working on a small docs change to PR locally and wanting to test it

PR #199, opened 2025-03-15
Updates for on potential false positives

AnthonyClark:
(Replace https://github.com/rspec/rspec/pull/198 - adds in deprecations)

1.
Updating the docs for `on_potential_false_positives`.

Before: the setter method does not show up in the docs due to how Yard handles attribute docs, and the attr_reader docs imply a truthy value.

After: More visibility into the values being used.

Screenshot 2025-03-11 at 9 11 32 AM

2.
Deprecates `warn_about_potential_false_positives` and `warn_about_potential_false_positives?` to enable removing them in a later version to simplify the codebase a bit.

### DIFF
--- a/rspec-expectations/lib/rspec/expectations/configuration.rb
+++ b/rspec-expectations/lib/rspec/expectations/configuration.rb
@@ -181,11 +181,20 @@ module RSpec
           # no-op, handler is something else
         end
       end
-      #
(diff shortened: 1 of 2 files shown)


pirj (approved):
Nice catch, thanks!

@JonRowe WDYT of only keeping `on_potential_false_positives` in 4.0?

PR #202, opened 2025-03-19
Remove backslashe comments in shared examples

baweaver:
When reading the [documentation for Shared Examples](https://rspec.info/features/3-13/rspec-core/example-groups/shared-examples/) I had noticed that some of the comments in code blocks had backslashes prepending them. This PR removes those backslashes.


pirj (approved):
Nice catch, thank you

JonRowe:
This is currently breaks the build, cucumber also has the concept of comments starting with `#` and I'm assuming that this is actually escaping the comments in the background section, potentially we could look at filtering them out during the doc building process but I'm on the fence about that as its not the best implementation currently...

PR #203, opened 2025-03-21
Add `detailed_message` detection to `RubyFeatures`

neilvcarvalho:
Adds a feature detection method to check if the Ruby version supports `Exception#detailed_message`, which was introduced in Ruby 3.2.0. This will enable future code to use the `detailed_message` method when available,  falling back to `message.to_s` on older Ruby versions.

### DIFF
--- a/rspec-support/lib/rspec/support/ruby_features.rb
+++ b/rspec-support/lib/rspec/support/ruby_features.rb
@@ -96,6 +96,16 @@ module RSpec
         end
       end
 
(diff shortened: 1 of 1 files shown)


neilvcarvalho:
As requested on https://github.com/rspec/rspec/pull/193/files#r2007228198

JonRowe:
Thanks, future updates to #193 should now not need us to manually accept the build :joy:

PR #205, opened 2025-04-23
Add GEM_HOME to exclusion patterns in backtrace formatter

pinzonjulian:
The base Docker Ruby image sets an environment variable called GEM_HOME to store gems in. With this change, the backtrace_formatter.rb is made aware of this location so it can also filter backtrace lines from gems within this location. This is particularly helpful when running code in CI platforms like Buildkite.

 You can see where this ENV variable is set in this link:

 https://github.com/docker-library/ruby/blob/235b3ffb2060c837137b32ea55f75816f2f4e4c4/3.4/bookworm/Dockerfile#L100C14-L100C31

### DIFF
--- a/rspec-core/lib/rspec/core/backtrace_formatter.rb
+++ b/rspec-core/lib/rspec/core/backtrace_formatter.rb
@@ -10,6 +10,7 @@ module RSpec
 
         patterns = %w[ /lib\d*/ruby/ bin/ exe/rspec /lib/bundler/ /exe/bundle: ]
         patterns << "org/jruby/" if RUBY_PLATFORM == 'java'
+        patterns << ENV["GEM_HOME"] if ENV.key?("GEM_HOME")
         patterns.map! { |s| Regexp.new(s.gsub("/", File::SEPARATOR)) }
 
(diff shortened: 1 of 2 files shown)


pinzonjulian:
I'm trying to run this spec locally but I'm having a hard time setting everything up. 
- I cloned the project
- ran bundle install
- ran a spec using `bundle exec rspec rspec-core/spec/rspec/core/backtrace_formatter_spec.rb` 
- Got this error`Could not find shared context "isolated directory"

I also tried some of the other instructions in DEVELOPMENT.md without any luck.

I'd appreciate it if someone could help me figure out how to run this to verify the change locally please. Thanks!

JonRowe:
:wave: Each "project" / gem is its own directory with its own test suite, you should be in the particular gems directory to run specs, by running from the root you are setting the working directory to the wrong place and thus support files aren't loaded. I'll look into modifying `development.md` to make this clearer.

However, at the moment I won't merge this, the default behaviour is to *not* exclude gem lines from backtraces, we provide the `filter_gems_from_backtrace` config to allow this if you want it, and that feature should work regardless of gem location as it matches on the gem name...

JonRowe:
I've tweaked the docs in #208

Issue #212, opened 2025-05-05
rspec_mocks 3.13.3 `and_call_original` breaks keyword args forwarding


timon:
`rspec_mocks` 3.13.3 breaks `and_call_original` for keyword arguments; 3.13.2 is just fine

I'm running this on ruby 3.3.6

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rspec-core', '~> 3.0', require: 'rspec/autorun'
  gem 'rspec-mocks', '3.13.3'
end

class RecursiveRunner
  def self.run(collection, delay:)
    new(collection).run(delay:)
  end

  def initialize(collection)
    @collection = collection
  end

  def run(delay:)
    @collection.shift
    if @collection.any?
      sleep(delay)
      run(delay:)
    end
  end
end

RSpec.describe RecursiveRunner do
  let(:collection) { %w(foo bar baz) }
  let(:delay) { 0 }

  describe 'a shorthand that does some recursion inside' do
    it 'does tail recursion with delay expected number of times' do
      expect_any_instance_of(described_class).to receive(:run).exactly(collection.size).times.and_call_original

      RecursiveRunner.run(collection, delay: delay)
    end
  end

  describe 'immediate recursive call' do
    subject(:runner) { described_class.new(collection) }

    it 'runs tail recursion' do
      expect(runner).to receive(:run).exactly(collection.size).times.and_call_original

      runner.run(delay: delay)
    end
  end
end
```

Expected outcome: all examples pass

Actual outcome: both fail with

```
ArgumentError:
       wrong number of arguments (given 1, expected 0; required keyword: delay)
```

noelrappin:
I'm seeing similar behavior, plus another one that I'm going to submit separately.

My issue was also with a nested call, and I think I isolated it to this commit https://github.com/rspec/rspec/pull/116

JonRowe:
#214 should fix this

Issue #213, opened 2025-05-05
RSpec Mocks 3.13.3 breaks nested mocks


noelrappin:
This is a little convoluted:

Test code:

```
class Sample
  attr_accessor :output

  def initialize(output: output)
    @output = output
  end

  def transaction(*args, **kwargs, &block)
    output.read_write do |connection|
      connection.transaction(*args, **kwargs, &block)
    end
  end

  def select(...)
    output.read_write do |connection|
      connection.select
    end
  end
end
```

Test

```
RSpec.describe(Sample) do
  context "double test" do
    let(:output) { double("output") }
    let(:proxy) { described_class.new(output: output) }

    it "does nested things" do
      primary = double(:connection)
      allow(primary).to receive(:transaction).and_yield
      allow(primary).to receive(:select)

      expect(output).to receive(:read_write).twice.and_yield(primary)

      proxy.transaction { proxy.select("true") }
    end
  end
end
```

This passes (correctly) in rspec_mocks 13.3.2 and fails in 13.3.3 with 

```
(Double "output").read_write(*(any args))
           expected: 2 times with any arguments
           received: 1 time with any arguments
```

I think it's because the inner nested call isn't counted, and I think I bisected it to this PR: https://github.com/rspec/rspec/pull/116

JonRowe:
#214 should fix this

PR #214, opened 2025-05-05
Allow nested mock calls by changing an internal safeguard

JonRowe:
#116 prevented nested calls from iterating counters to defend against recursion during error message formatting, but this can break legitimate nested loops, so we switch that to only safeguarding internals e.g the error generation against this.


JonRowe:
Fixes #212 #213

orangewolf:
@JonRowe thanks for the speedy fix for this!

JonRowe:
Released in 3.13.4

PR #219, opened 2025-05-23
Clear thread local internal variable rather than set to false fixes #218

### LINKED ISSUE #218: Memory leak in rspec-mocks 3.13.3
When upgrading rspec-mocks from 3.13.2, we were getting consistent CI failures where the rspec process would run out of memory. Some basic `ps` monitoring showed that memory usage would slowly leak over the course of our large suite: no sharp all-at-once increase, just a gradual slope monotonically increasing as each test ran until eventually we hit our CI limits.

Investigating the diffs, I think the culprit is #156, where a new thread-local variable gets stored for each separate expectation for rspec mocks. As long as the thread lives the variables won't get garbage-collected, even though (afaict) their values only matter for a short period of time.

### Demonstration

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rspec', '~> 3'
  gem 'rspec-mocks', ENV['RSPEC_MOCKS']
  gem 'get_process_mem'
end

require 'rspec/autorun'

def report_growth_factor(thing, before, after)
  percentage = (after - before) / before.to_f
  human_percentage = (100 * percentage).round(2)
  puts "#{thing} grew from #{before} to #{after} (#{human_percentage}%)"
end

RSpec.configure do |config|
  config.before(:suite) do
    $memory_before = GetProcessMem.new(Process.pid).bytes
    $thread_before = RSpec::Support.thread_local_data.size
  end

  config.after(:suite) do
    $memory_after = GetProcessMem.new(Process.pid).bytes
    $thread_after = RSpec::Support.thread_local_data.size
    puts "\n\n"
    report_growth_factor('Bytes of memory', $memory_before, $memory_after)
    report_growth_factor('Number of thread-local variables', $thread_before, $thread_after)
  end
end

class Mock
  def message
    :ok
  end
end

RSpec.describe 'memory leak' do
  ENV['N'].to_i.times do
    it 'uses an rspec mock' do
      mock = Mock.new
      expect(mock).to receive(:message).once.and_call_original
      expect(mock.message).to be :ok
    end
  end
end
```

```console
$ ruby --version
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-darwin22]
```

```console
$ N=100000 RSPEC_MOCKS=3.13.2 ruby /tmp/leak.rb
[...]

Bytes of memory grew from 391946240 to 426946560 (8.93%)
Number of thread-local variables grew from 2 to 3 (50.0%)


Finished in 17.13 seconds (files took 2.05 seconds to load)
100000 examples, 0 failures
```

```console
$ N=100000 RSPEC_MOCKS=3.13.3 ruby /tmp/leak.rb
[...]

Bytes of memory grew from 393396224 to 464261120 (18.01%)
Number of thread-local variables grew from 2 to 100003 (5000050.0%)


Finished in 18.29 seconds (files took 2 seconds to load)
100000 examples, 0 failures
```

JonRowe:
#156 inadvertently leads to a growing amount thread variables as the key is object specific and won't be cleaned up until the thread dies which on MRI is never? Instead set to true when setting, then delete the key when setting to false.

### DIFF
--- a/rspec-mocks/lib/rspec/mocks/message_expectation.rb
+++ b/rspec-mocks/lib/rspec/mocks/message_expectation.rb
@@ -620,7 +620,14 @@ module RSpec
         end
 
         def invoking_internals=(value)
-          RSpec::Support.thread_local_data[:"__rspec_#{object_id}_invoking_internals"] = value
+          # We clear the key for this rather than setting to false because otherwise the amount of
+          # thread local data will keep growing over the lifetime of the thread (which is a long
+          # time on a single threaded spec run e.g standard MRI)
+          if value
+            RSpec::Support.thread_local_data[:"__rspec_#{object_id}_invoking_internals"] = true
+          else
+            RSpec::Support.thread_local_data.delete(:"__rspec_#{object_id}_invoking_internals")
+          end
         end
 
         def invoke_incrementing_actual_calls_by(increment, allowed_to_fail, parent_stub, *args, &block)


benoittgt (approved):
Should we add a non-regression test based on the initial bug report?

JonRowe:
> Should we add a non-regression test based on the initial bug report?

I'm open to adding better measurement around this but I feel like a specific regression test would either be too complicated to implement or too specific and brittle. We've done previous benchmarking around this its just not a CI safe guard, it would be good to have that but the only specific regression test you could add here easily would be to check that you don't leave the variable around which would be brittle.

PR #240, opened 2025-08-21
Fix `World::Null.traverse_example_group_trees_until` to accept a dummy block argument

viralpraxis:
To fix this warning

```
rspec-core-3.13.5/lib/rspec/core/configuration.rb:2370: warning: the block passed to 'RSpec::Core::World::Null.traverse_example_group_trees_until' defined 3.4.0/gems/rspec-core-3.13.5/lib/rspec/core/world.rb:272 may be ignored
```

we can simply change `Null.traverse_example_group_trees_until` to accept a `&_block`, just like `World#traverse_example_group_trees_until` does.

### DIFF
--- a/rspec-core/Changelog.md
+++ b/rspec-core/Changelog.md
@@ -7,6 +7,11 @@ Enhancements:
   allways print a line number rather than an example id when the line number is ambiguous.
   (Baden Ashford, rspec/rspec-core#3085)
 
+Bug Fixes:
+
+* Add explicit block parameter to `RSpec::World::Null.traverse_example_group_trees_until` to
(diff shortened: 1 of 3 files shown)


JonRowe:
How were you generating this warning? I've never seen it and our test suite checks for warnings.

viralpraxis:
> How were you generating this warning? I've never seen it and our test suite checks for warnings.

That's a good question. Initially, I saw it during running my app's test suite, but for some reason I can't repro it anymore. Perhaps it is a flaky warning

Anyway, I think this patch makes sense. Null implementation should have the same method signatures as the main one, right?

JonRowe:
Released in 3.13.6

PR #243, opened 2025-09-04
Warn when include_examples is used with a block

### LINKED ISSUE #230: Warn when using include_examples with a block
### Background

`include_examples` does not create a wrapping example group, and is a "noisy neighbour" when used in the same example group with other elements, and can introduce subtle issues.
Our [RSpec Core's readme](https://github.com/rspec/rspec/blob/95fbf88255de04e4ba8288172292bda494c4f187/rspec-core/README.md?plain=1#L77) only mentions `include_examples`, and doesn't even mention a more reliable `it_behaves_like` alternative, that does create a wrapping example group, that doesn't let hooks, `let` definitions and method definitions to litter the including example group.

### Problem

The problem I want to highlight in this issue is narrow, it's the case when a block is passed to `include_examples`.

```ruby
    describe do
      shared_examples "equals foo" do
        it do
          expect(v).to eq("foo")
        end
      end

      include_examples "equals foo" do
        let(:v) { "bar" } # should 💥, but it won't
      end

      include_examples "equals foo" do
        let(:v) { "foo" }
      end
    end
```
Both those examples will pass, and will both fail if you swap the order.

It produces deceitful specs, it looks like the definitions inside the block are isolated, but they are not.
Not only `let`s leak, but also hooks and method declarations:
```ruby
      include_examples "equals foo" do
        before { use("include_examples").everywhere }
        def hoping_it_works_intuitively
          "looking aside"
        end
      end
```

### Real world motivation

Real-world [usages](https://github.com/search?q=%2Finclude_examples.%2B+do%2F+lang%3Aruby+path%3A%2Fspec%5C%2F%2F&type=code): [1](https://github.com/rubycdp/ferrum/blob/180f29297f2be7b770b32d8a2f73343e9530e67c/spec/browser_spec.rb#L418), [2](https://github.com/Mange/roadie-rails/blob/117b2567ad6aa4cbf1d1fc0a470f4f0ac756e2db/spec/integration_spec.rb#L56), [3](https://github.com/rubycdp/ferrum/blob/180f29297f2be7b770b32d8a2f73343e9530e67c/spec/browser_spec.rb#L418), [4](https://github.com/Mange/roadie-rails/blob/117b2567ad6aa4cbf1d1fc0a470f4f0ac756e2db/spec/integration_spec.rb#L56) and many more.

### Proposal

1. Introduce a warning in 3.x (or 3.99) when `include_examples` is used with a block, suggest using `it_behaves_like` instead.
2. Replace `include_examples` with `it_behaves_like` in README
3. Fail hard if a block is passed to `include_examples` in 4.0

### Misc

[Discussion on community style guide](https://github.com/rubocop/rspec-style-guide/issues/66).

pirj:
(partially) fixes https://github.com/rspec/rspec/issues/230
The 4.0 counterpart https://github.com/rspec/rspec/pull/259

:question: Open questions: should we also warn when it is used with arguments?
We warn about this in [our documentation](https://rspec.info/features/3-13/rspec-core/example-groups/shared-examples/):
> When you include parameterized examples in the current context multiple times, you may override previous method definitions and last declaration wins.

### DIFF
--- a/rspec-core/README.md
+++ b/rspec-core/README.md
@@ -89,7 +89,10 @@ You can declare examples within a group using any of `it`, `specify`, or
 ## Shared Examples and Contexts
 
 Declare a shared example group using `shared_examples`, and then include it
-in any group using `include_examples`.
+in any group using `include_examples`, or `it_behaves_like`.
+
+Note that `include_examples` directly includes examples into the current context so
+you often want to use `it_behaves_like` to isolate the examples into a context.
 
 ```ruby
 RSpec.shared_examples "collections" do |collection_class|
@@ -103,7 +106,7 @@ RSpec.describe Array do
 end
 
 RSpec.describe Hash do
-  include_examples "collections", Hash
+  it_behaves_like "collections", Hash
 end
 ```
 
@@ -164,11 +167,11 @@ RSpec.shared_examples "collections" do
 end
 
 RSpec.describe Array do
-  include_examples "collections"
+  it_behaves_like "collections"
 end
 
 RSpec.describe Hash do
-  include_examples "collections"
+  it_behaves_like "collections"
 end
 ```
 
--- a/rspec-core/features/example_groups/shared_examples.feature
+++ b/rspec-core/features/example_groups/shared_examples.feature
@@ -58,9 +58,7 @@ Feature: Using shared examples
   end
   ```
 
-  To prevent this kind of subtle error a warning is emitted if you declare multiple
-  methods with the same name in the same context. Should you get this warning
-  the simplest solution is to replace `include_examples` with `it_behaves_like`, in this
+  The simplest solution is to replace `include_examples` with `it_behaves_like`, in this
   way method overriding is avoided because of the nested context created by `it_behaves_like`
 
   Conventions:
--- a/rspec-core/lib/rspec/core/example_group.rb
+++ b/rspec-core/lib/rspec/core/example_group.rb
@@ -362,6 +362,10 @@ module RSpec
       #
       # @see SharedExampleGroup
       def self.include_examples(name, *args, &block)
+        if block
+          RSpec.deprecate("Passing a block to `include_examples`",
+                          :replacement => "Use `it_behaves_like` instead")
+        end
         find_and_eval_shared("examples", name, caller.first, *args, &block)
       end
 
--- a/rspec-core/lib/rspec/core/example_group.rb
+++ b/rspec-core/lib/rspec/core/example_group.rb
@@ -352,6 +352,13 @@ module RSpec
       #
       # @see SharedExampleGroup
       def self.include_context(name, *args, &block)
+        if block
+          RSpec.deprecate(
+            "Passing a block to `include_context`",
+            :replacement =>
+              "Use `it_behaves_like` instead, or place the block content " \
+              "after the statement.")
+        end
         find_and_eval_shared("context", name, caller.first, *args, &block)
       end
 
@@ -363,8 +370,11 @@ module RSpec
       # @see SharedExampleGroup
       def self.include_examples(name, *args, &block)
         if block
-          RSpec.deprecate("Passing a block to `include_examples`",
-                          :replacement => "Use `it_behaves_like` instead")
+          RSpec.deprecate(
+            "Passing a block to `include_examples`",
+            :replacement =>
+              "Use `it_behaves_like` instead, or place the block content " \
+              "after the statement.")
         end
         find_and_eval_shared("examples", name, caller.first, *args, &block)
       end
--- a/rspec-core/lib/rspec/core/example_group.rb
+++ b/rspec-core/lib/rspec/core/example_group.rb
@@ -356,8 +356,9 @@ module RSpec
           RSpec.deprecate(
             "Passing a block to `include_context`",
             :replacement =>
-              "Use `it_behaves_like` instead, or place the block content " \
-              "after the statement.")
+              "Either use `it_behaves_like` to wrap the block " \
+              "contents in a context, or place the block content " \
+              "within the parent context.")
         end
         find_and_eval_shared("context", name, caller.first, *args, &block)
       end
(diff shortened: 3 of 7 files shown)


rspec-core/features/example_groups/shared_examples.feature
```diff
@@ -58,9 +58,7 @@ Feature: Using shared examples
   end
   ```
 
-  To prevent this kind of subtle error a warning is emitted if you declare multiple
```
pirj, line 61: That old warning existed, introduced [here](https://github.com/rspec/rspec-core/commit/e072e5358b85a971010bc1a22d04eeca646321c8#diff-e779aa0d3ee19a0e32fad4a02f2f12e28fa4ab78befa68ae10fab8a0eb06194bR658), but soon reverted [here](https://github.com/rspec/rspec-core/commit/b6630284):
> The warning is overzealous. I got spammed with tons of warnings of situations that weren't actually problematic

rspec-core/spec/rspec/core/dsl_spec.rb
```diff
@@ -54,7 +54,7 @@ def disable
   end
 
   describe "built in DSL methods" do
-    include_examples "dsl methods", :describe, :context, :shared_examples, :shared_examples_for, :shared_context do
+    it_behaves_like "dsl methods", :describe, :context, :shared_examples, :shared_examples_for, :shared_context do
```
JonRowe: Rather than nest contexts here, change the example to not pass the block and define the method directly in the outer describe for all of these.

rspec-core/README.md
```diff
@@ -87,11 +87,11 @@ RSpec.shared_examples "collections" do |collection_class|
 end
 
 RSpec.describe Array do
-  include_examples "collections", Array
+  it_behaves_like "collections", Array
 end
 
 RSpec.describe Hash do
-  include_examples "collections", Hash
+  it_behaves_like "collections", Hash
```
JonRowe, line 109: Add an example of each type.

rspec-core/README.md
```diff
@@ -77,7 +77,7 @@ You can declare examples within a group using any of `it`, `specify`, or
 ## Shared Examples and Contexts
 
 Declare a shared example group using `shared_examples`, and then include it
-in any group using `include_examples`.
+in any group using `it_behaves_like`.
```
JonRowe:
```suggestion
in any group using `include_examples`, or `it_behaves_like`. 

Note that `include_examples` directly includes examples into the current context so
you often want to use `it_behaves_like` to isolate the examples into a context.
```

rspec-core/README.md
```diff
@@ -100,7 +100,7 @@ within a shared example group. This includes `before`, `after`, and `around`
 hooks, `let` declarations, and nested groups/contexts.
 
 You can also use the names `shared_context` and `include_context`. These are
-pretty much the same as `shared_examples` and `include_examples`, providing
+pretty much the same as `shared_examples` and `it_behaves_like`, providing
```
JonRowe: Revert this.

rspec-core/spec/rspec/core/dsl_spec.rb
```diff
@@ -63,7 +63,7 @@ def changing_expose_dsl_globally
 
   describe "custom example group aliases" do
     context "when adding aliases before exposing the DSL globally" do
-      include_examples "dsl methods", :detail do
+      it_behaves_like "dsl methods", :detail do
```
JonRowe: Drop the block here and below.

JonRowe (changes requested):
So l I have no plans to release another minor (other than 3.99) before 4, thus this can make a breaking change, with another PR to 3.99 to add the deprecation, stylistically I have some comments about the text changes but other than that this is a good idea.

rspec-core/README.md
pirj:
```suggestion
pretty much the same as `shared_examples` and `include_examples`, providing
```

rspec-core/lib/rspec/core/example_group.rb
```diff
@@ -358,6 +358,10 @@ def self.include_context(name, *args, &block)
       #
       # @see SharedExampleGroup
       def self.include_examples(name, *args, &block)
+        if block
+          RSpec.deprecate("Passing a block to `include_examples`",
+                          :replacement => "Use `it_behaves_like` instead")
```
pirj: @JonRowe do you plan to merge this PR to the future 4.0 branch, or to 3.99 branch?
3.99 needs the `RSpec.deprecate`, while 4.0 will have a `raise` here, right?
```suggestion
          raise "Passing a block to `include_examples` is not allowed. " \
                 "Use `it_behaves_like` instead."
```

JonRowe: `main` is now 4.0.0.pre, so I'd appreciate a raise here and a seperate pr adding a deprecation to 3-99-`maintenance`

pirj:
@JonRowe I intend to extend this to `include_context` just as well, as
```ruby
include_context "foo" do
  let(:bar) { "baz" }
end
```
is the same as the "flat form":
```ruby
include_context "foo"
let(:bar) { "baz" }
```

I acknowledge that semantically `include_context` must serve a different purpose, but practically they are freely used interchangeably, and while we warn/raise to encourage people to prefer the "flat form", or `it_behaves_like` to `include_examples`, we would be leaving the backdoor to use `include_context`.

JonRowe:
I've left feedback about some style / docs stuff on the alternate PR that also applies here FWIW

pirj:
@JonRowe Just a heads-up. GitHub closed this, apparently due to this:

image

JonRowe:
Yeah I accidentally deleted the branch when rebasing (its not protected because its essentially a dev branch until theres a release or beta) need to check there aren't any more closed PRs, I'll probably setup a branch rule now :joy:

pirj:
I really hope nobody will attempt to run 3.99 with JRuby 1.7 🙈
The same failure affected https://github.com/rspec/rspec/pull/107, might be unrelated to the PR?

Otherwise, green.

pirj:
Pushed one last commit to align the deprecation warning text to [the error text is in 4.0](https://github.com/rspec/rspec/pull/259#discussion_r2394932688).

JonRowe:
Manually merged to `3-99-maintenance`

PR #246, opened 2025-09-12
Restore the deleted simplecov script

pirj:
[-f ] check was effectively skipping it


JonRowe:
Only rspec-core needs the script because its a special case, thats why the check exists and the file exists in the rspec-core repo, the other repos shouldn't need it.

PR #248, opened 2025-09-14
Fix annoying flakiness

pirj:
Ex https://github.com/rspec/rspec/actions/runs/17475121455/job/49632938114

To reproduce:
```
cd rspec-support
rspec --seed 50557
```

```
Failures:

  1) Differ #diff handles any encoding error that occurs with a helpful error message
     Failure/Error:
       expect(RSpec::Support::HunkGenerator).to receive(:new).
         and_raise(Encoding::CompatibilityError)

     NameError:
       uninitialized constant RSpec::Support::HunkGenerator
     # ./spec/rspec/support/differ_spec.rb:170:in `block (3 levels) in <module:Support>'
```

Introduced [here](https://github.com/rspec/rspec-support/pull/599/files#diff-2106e669e7c9f2c1b8d72795c4180615598160d95ac05648ed78a96d11ada0e6R9)

### DIFF
--- a/rspec-support/spec/rspec/support/differ_spec.rb
+++ b/rspec-support/spec/rspec/support/differ_spec.rb
@@ -7,7 +7,7 @@ require 'rspec/support/spec/string_matcher'
 
 module RSpec
   module Support
-    RSpec.describe "Differ" do
+    RSpec.describe Differ do
       include Spec::DiffHelpers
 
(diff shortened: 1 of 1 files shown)


JonRowe:
I think the better solution is to autoload the hunk generator, see #249

PR #252, opened 2025-09-26
Raise ArgumentError when docstring's are not a string (or nil)

### LINKED ISSUE #40: doc_string which is the first argument of example, must be string object

JonRowe:
This was originallly rspec/rspec-core#3081 it got merged prior to the monorepo migration to get it into 4-0-dev but it has some consquences when run across the entire build surrounding metadata, and needs properly allowing classes (even if I have a preference against them, they are too widly used to "break" in 4 plus theres the surrounding debate about `described_class` we've yet to have) in example groups.

3.99 counterpart https://github.com/rspec/rspec/pull/107

Fixes https://github.com/rspec/rspec/issues/40

### DIFF
--- a/rspec-core/Changelog.md
+++ b/rspec-core/Changelog.md
@@ -35,6 +35,10 @@ Breaking Changes:
   (Phil Pirozhkov, rspec/rspec-core#2864)
 * Unify multi-condition filtering to use "all" semantic. (Phil Pirozhkov, rspec/rspec-core#2874)
 * Change the default order to random. (Santiago Bartesaghi, rspec/rspec-core#2929)
+* Raise an error when an example group description is neither a string, class or module.
+  (Eisuke Yoshida, rspec/rspec-core#3081)
+* Raise an error when an example description is nor a string.
+  (Eisuke Yoshida, rspec/rspec-core#3081)
 
 Enhancements:
 
--- a/rspec-core/lib/rspec/core/example_group.rb
+++ b/rspec-core/lib/rspec/core/example_group.rb
@@ -146,6 +146,10 @@ module RSpec
         idempotently_define_singleton_method(name) do |*all_args, &block|
           desc, *args = *all_args
 
+          unless NilClass === desc || String === desc
+            raise ArgumentError, "Examples must be described with a string, got: `#{desc.inspect}`"
+          end
+
           options = Metadata.build_hash_from(args)
           options.update(:skip => RSpec::Core::Pending::NOT_YET_IMPLEMENTED) unless block
           options.update(extra_options)
@@ -269,6 +273,10 @@ module RSpec
             combined_metadata = metadata.dup
             combined_metadata.merge!(args.pop) if args.last.is_a? Hash
             args << combined_metadata
+
+            unless NilClass === description || String === description || Class === description || Module === description
+              raise ArgumentError, "Example groups must be described with a string, got: `#{description.inspect}`"
+            end
 
             subclass(self, description, args, registration_collection, &example_group_block)
           ensure
--- a/rspec-core/spec/rspec/core/example_group_spec.rb
+++ b/rspec-core/spec/rspec/core/example_group_spec.rb
@@ -55,16 +55,6 @@ module RSpec::Core
       end
     end
 
-    it 'does not treat the first argument as a metadata key even if it is a symbol' do
-      group = RSpec.describe(:symbol)
-      expect(group.metadata).not_to include(:symbol)
-    end
-
-    it 'treats the first argument as part of the description when it is a symbol' do
-      group = RSpec.describe(:symbol)
-      expect(group.description).to eq("symbol")
-    end
-
     describe "constant naming" do
       around do |ex|
         before_constants = RSpec::ExampleGroups.constants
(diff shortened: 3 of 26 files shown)


rspec-core/Changelog.md
```diff
@@ -35,6 +35,10 @@ Breaking Changes:
   (Phil Pirozhkov, rspec/rspec-core#2864)
 * Unify multi-condition filtering to use "all" semantic. (Phil Pirozhkov, rspec/rspec-core#2874)
 * Change the default order to random. (Santiago Bartesaghi, rspec/rspec-core#2929)
+* Raise an error when an example group description is neither a string, class or module.
+  (Eisuke Yoshida, rspec/rspec-core#3081)
+* Raise an error when an example description is nor a string.
```
pirj, line 40:
```suggestion
* Raise an error when an example description is not a string nor a class.
```

pirj:
Green

pirj:
Green, squashed.

rspec-core/lib/rspec/core/example_group.rb
```diff
@@ -270,6 +274,10 @@ def self.define_example_group_method(name, metadata={})
             combined_metadata.merge!(args.pop) if args.last.is_a? Hash
             args << combined_metadata
 
+            unless NilClass === description || String === description || Class === description || Module === description
+              raise ArgumentError, "Example groups must be described with a string, got: `#{description.inspect}`"
```
JonRowe, line 278:
```suggestion
              raise ArgumentError, "Example groups must be described with a string, class or module got: `#{description.inspect}`"
```

rspec-core/spec/rspec/core/example_group_spec.rb
```diff
@@ -55,16 +55,6 @@ def metadata_hash(*args)
       end
     end
 
-    it 'does not treat the first argument as a metadata key even if it is a symbol' do
-      group = RSpec.describe(:symbol)
-      expect(group.metadata).not_to include(:symbol)
-    end
```
JonRowe, line 61: These should probably be treated as metadata...

rspec-core/spec/rspec/core/example_group_spec.rb
```diff
@@ -1233,6 +1223,66 @@ def extract_execution_results(group)
...
+
+      it "raises ArgumentError when a Symbol is used as an example group doc string" do
+        expect { RSpec.describe :foo }.
+          to raise_error(ArgumentError, /Example groups must be described with a string, got: `:foo`/)
+      end
+
+      it "raises ArgumentError when a Hash is used as an example group doc string" do
+        expect { RSpec.describe(foo: :bar) { } }.
+          to raise_error(%r{Example groups must be described with a string, got: `#{Regexp.escape({ foo: :bar }.inspect)}`})
+      end
```
JonRowe, line 1251: No this is incorrect, this should be interepreted as a `nil` description with metadata, I think that should still be supported for now, just because their are a tonne of aliases and things where that makes sense.

Think `describe type: :model { ... }` for example.

rspec-core/spec/rspec/core/example_group_spec.rb
```diff
@@ -1233,6 +1223,66 @@ def extract_execution_results(group)
...
+
+      it "raises ArgumentError when a Symbol is used as an example doc string" do
+        expect { group.it(:foo) { } }.
+          to raise_error(ArgumentError, /Examples must be described with a string, got: `:foo`/)
+      end
+
+      it "raises ArgumentError when a Hash is used as an example doc string" do
+        expect { group.it(foo: :bar) { } }.
+          to raise_error(ArgumentError, %r{Examples must be described with a string, got: `#{Regexp.escape({ foo: :bar }.inspect)}`})
```
JonRowe, line 1282: Similarly, this is a `nil` description with metadata, this will be more common...
```
specify(tag: :validations) { is_expected.to be_valid }
```

rspec-core/spec/rspec/core/example_group_spec.rb
```diff
@@ -1233,6 +1223,66 @@ def extract_execution_results(group)
...
+      end
+
+      it "accepts example group without a doc string" do
+        expect { RSpec.describe }.not_to output.to_stderr
+      end
+
+      it "raises ArgumentError when a Symbol is used as an example group doc string" do
+        expect { RSpec.describe :foo }.
+          to raise_error(ArgumentError, /Example groups must be described with a string, got: `:foo`/)
+      end
```
JonRowe, line 1246: Technically `:foo` here should be considered `foo: true` and be valid as metadata.

JonRowe (commented):
Thanks for your work here @pirj but this still doesn't tackle the bare "metadata" cases, see comments.

pirj:
I see what you mean, @JonRowe

One thing is that any symbol can be treated as metadata, while in most cases it’s not.
Sometimes, it will be harmless, but when this matches some metadata trigger, it can cause surprising behaviour.

In the same way, since most Ruby hashes have symbol keys anyway, and since the same is our deciding factor if it is metadata or not, we can mistakenly take some hash from the first argument and treat it as metadata, while it is not.

I’m aways from my computer and can’t take a closer look now.
However, from memory, even thiugh I would like to formalise this so the first argument should always be a string (class, module). But if we’d introduce some exceptions to this rule, and add detection by type, I would rather not do this at all and kept things as they are.

PR #259, opened 2025-09-26
Raise on include_examples/context with a block

### LINKED ISSUE #230: Warn when using include_examples with a block

pirj:
When you include parameterized examples in the current context multiple times, you may override previous method definitions and last declaration wins.

Fixes #230

Targets 4.0-pre
For 3.99, there's #243

### DIFF
--- a/rspec-core/README.md
+++ b/rspec-core/README.md
@@ -75,7 +75,7 @@ You can declare examples within a group using any of `it`, `specify`, or
 ## Shared Examples and Contexts
 
 Declare a shared example group using `shared_examples`, and then include it
-in any group using `include_examples`.
+in any group using `it_behaves_like` or `include_examples`.
 
 ```ruby
 RSpec.shared_examples "collections" do |collection_class|
@@ -84,14 +84,38 @@ RSpec.shared_examples "collections" do |collection_class|
   end
 end
 
+RSpec.shared_examples "implements #empty?" do
+  it "returns true when there are no members" do
+    expect(empty_target).to be_empty
+  end
+
+  it "returns false when there are members" do
+    expect(non_empty_target).to_not be_empty
+  end
+end
+
 RSpec.describe Array do
   include_examples "collections", Array
+
+  it_behaves_like "implements #empty?" do
+    let(:empty_target) { Array.new }
+    let(:non_empty_target) { [:key] }
+  end
 end
 
 RSpec.describe Hash do
   include_examples "collections", Hash
+
+  it_behaves_like "implements #empty?" do
+    let(:empty_target) { Hash.new }
+    let(:non_empty_target) { {key: :value} }
+  end
 end
 ```
+
+Note that `include_examples` directly includes examples into the current context, whilst
+`it_behaves_like` wraps the examples in a new context which allows you to pass a block
+to customise things like `let` or `subject` without using local variables passed in.
 
 Nearly anything that can be declared within an example group can be declared
 within a shared example group. This includes `before`, `after`, and `around`
--- a/rspec-core/lib/rspec/core/example_group.rb
+++ b/rspec-core/lib/rspec/core/example_group.rb
@@ -344,8 +344,10 @@ module RSpec
       # context.
       #
       # @see SharedExampleGroup
-      def self.include_context(name, *args, &block)
-        find_and_eval_shared("context", name, caller.first, *args, &block)
+      def self.include_context(name, *args)
+        issue_block_inclusion_error!("include_context") if block_given?
+
+        find_and_eval_shared("context", name, caller.first, *args)
       end
 
       # Includes shared content mapped to `name` directly in the group in which
@@ -354,8 +356,10 @@ module RSpec
       # context.
       #
       # @see SharedExampleGroup
-      def self.include_examples(name, *args, &block)
-        find_and_eval_shared("examples", name, caller.first, *args, &block)
+      def self.include_examples(name, *args)
+        issue_block_inclusion_error!("include_examples") if block_given?
+
+        find_and_eval_shared("examples", name, caller.first, *args)
       end
 
       # Clear memoized values when adding/removing examples
@@ -391,6 +395,15 @@ module RSpec
           self, Metadata.relative_path(inclusion_location),
           args, customization_block
         )
+      end
+
+      # @private
+      def self.issue_block_inclusion_error!(method)
+        raise ArgumentError,
+              "Do not pass a block to `#{method}`. " \
+              "Either use `it_behaves_like` to wrap the block " \
+              "contents in a context, or place the block content " \
+              "within the parent context."
       end
 
       # @!endgroup
--- a/rspec-core/spec/rspec/core/dsl_spec.rb
+++ b/rspec-core/spec/rspec/core/dsl_spec.rb
@@ -20,18 +20,18 @@ RSpec.describe "The RSpec DSL" do
   end
 
   describe "built in DSL methods" do
-    include_examples "dsl methods", :describe, :context, :shared_examples, :shared_examples_for, :shared_context do
-      def setup
-      end
+    include_examples "dsl methods", :describe, :context, :shared_examples, :shared_examples_for, :shared_context
+
+    def setup
     end
   end
 
   describe "custom example group aliases" do
     context "when adding aliases before exposing the DSL globally" do
-      include_examples "dsl methods", :detail do
-        def setup
-          RSpec.configuration.alias_example_group_to(:detail)
-        end
+      include_examples "dsl methods", :detail
+
+      def setup
+        RSpec.configuration.alias_example_group_to(:detail)
       end
     end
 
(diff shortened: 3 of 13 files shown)


pirj:
@JonRowe Both this and #243 are green. Changes since you last checked:
 - added `include_context` ([reasoning](https://github.com/rspec/rspec/pull/243#issuecomment-3340495670))
 - changed text to propose two resolutions: "flat form" or `it_behaves_like`

rspec-core/lib/rspec/core/example_group.rb
```diff
@@ -344,8 +344,14 @@ def self.define_nested_shared_group_method(new_name, report_label="behaves like"
...
       #
       # @see SharedExampleGroup
-      def self.include_context(name, *args, &block)
-        find_and_eval_shared("context", name, caller.first, *args, &block)
+      def self.include_context(name, *args)
+        if block_given?
+          raise ArgumentError,
+                "Don't pass a block to `include_context`. " \
+                "Use `it_behaves_like` instead, or place the block content " \
+                "after the statement."
```
pirj: `it_behaves_like` is not a proper replacement for a correctly semantically used `include_context`, should we instead just
```suggestion
                "Don't pass a block to `include_context`. " \
                "Place the block content after the statement."
```

rspec-core/spec/rspec/core/dsl_spec.rb
```diff
@@ -20,19 +20,17 @@
   end
 
   describe "built in DSL methods" do
-    include_examples "dsl methods", :describe, :context, :shared_examples, :shared_examples_for, :shared_context do
-      def setup
-      end
+    def setup
     end
+    include_examples "dsl methods", :describe, :context, :shared_examples, :shared_examples_for, :shared_context
```
JonRowe:
```suggestion
    include_examples "dsl methods", :describe, :context, :shared_examples, :shared_examples_for, :shared_context

    def setup
    end
```

rspec-core/spec/rspec/core/dsl_spec.rb
```diff
@@ -20,19 +20,17 @@
...
   describe "custom example group aliases" do
     context "when adding aliases before exposing the DSL globally" do
-      include_examples "dsl methods", :detail do
-        def setup
-          RSpec.configuration.alias_example_group_to(:detail)
-        end
+      def setup
+        RSpec.configuration.alias_example_group_to(:detail)
       end
+      include_examples "dsl methods", :detail
```
JonRowe:
```suggestion
      include_examples "dsl methods", :detail

      def setup
        RSpec.configuration.alias_example_group_to(:detail)
      end
```

rspec-mocks/spec/rspec/mocks/matchers/have_received_spec.rb
```diff
@@ -673,12 +673,11 @@ def double_with_unmet_expectation(method_name)
     end
 
     RSpec.describe Matchers::HaveReceived, "when used in a context that has rspec-mocks and rspec-expectations available" do
-      include_examples Matchers::HaveReceived do
-        # Override `fail_including` for this context, since `have_received` is a normal
-        # rspec-expectations matcher, the error class is different.
-        def fail_including(*snippets)
-          raise_error(RSpec::Expectations::ExpectationNotMetError, a_string_including(*snippets))
-        end
+      include_examples Matchers::HaveReceived
```
JonRowe, line 676:
```suggestion
      include_examples Matchers::HaveReceived

```

rspec-core/spec/rspec/core/formatters/exception_presenter_spec.rb
```diff
@@ -828,14 +828,18 @@ def exception_with(backtrace)
     end
   end
 
-  RSpec.shared_examples_for "a class satisfying the common multiple exception error interface" do
+  RSpec.shared_context "exception helpers" do
```
JonRowe, line 821: It doesn't look like this is necessary, if you kept this as it was and just changed to `it_behaves_like` I feel like it should work, if it doesn't why not?

rspec-core/README.md
```diff
@@ -89,7 +92,7 @@ RSpec.describe Array do
 end
 
 RSpec.describe Hash do
-  include_examples "collections", Hash
+  it_behaves_like "collections", Hash
```
JonRowe: I feel like some clarifying docs / expanding the example to highlight the difference is needed.

rspec-core/lib/rspec/core/example_group.rb
```diff
@@ -344,8 +344,14 @@ def self.define_nested_shared_group_method(new_name, report_label="behaves like"
...
+        end
+        find_and_eval_shared("context", name, caller.first, *args)
```
JonRowe:
```suggestion
        if block_given?
          raise ArgumentError,
                "Don't pass a block to `include_context`. " \
                "Either use `it_behaves_like` to wrap the block " \
                "contents in a context, or place the block content " \
                "within the parent context."
        end

        find_and_eval_shared("context", name, caller.first, *args)
```

rspec-core/lib/rspec/core/example_group.rb
JonRowe: No because sometimes you actually want `it_behaves_like` and the usage was just broken, see suggestion below.

rspec-core/spec/rspec/core/example_group_spec.rb
```diff
@@ -1719,20 +1720,35 @@ def foo; end
...
+
+        context "when not passed a block" do
+          it "does not raise" do
+            group = RSpec.describe("host group") do
+              shared_examples "some behavior" do
+                example "shared example"
+              end
             end
+
+            group.public_send(name, "some behavior")
```
JonRowe: You can remove this context/test its already covered above and we don't need a specific example.

rspec-core/spec/rspec/core/example_group_spec.rb
```diff
@@ -1719,20 +1720,35 @@ def foo; end
...
+            allow(Warning).to receive(:warn).with(/the block passed .+ may be ignored/)
+
+            expect {
+              group.public_send(name, "some behavior") do
+                let(:leak) { "boom" }
+              end
+            }.to raise_error(
+              "Don't pass a block to `#{name}`. Use `it_behaves_like` instead, or place the block content after the statement.")
+          end
+        end
```
JonRowe: Flatten the context out, you only have one spec and the context doesn't add anything.

JonRowe (changes requested):
Getting there, just some tweaks to formatting and specs plus a question

rspec-core/spec/rspec/core/formatters/exception_presenter_spec.rb
pirj: Good question.
There are no examples in this context, only two helper methods. The previous docstring deceptively suggested that including this may include some specs related to the multiple error interface, but it didn't.

PR #265, opened 2025-10-12
Avoid infinite loop when classes mock #is_a?

### LINKED ISSUE #73: `stack level too deep` when stubbing :is_a? with allow_any_instance_of
### Subject of the issue
The error "`stack level too deep`" occurs when stubbing `:is_a?` with `allow_any_instance_of`.

### Your environment
* Ruby version: ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [x86_64-linux]

* rspec-mocks version: 3.12.6

##### stub_is_a_spec.rb
```ruby
describe '`stack level too deep` when stubbing :is_a? ' do
  let(:klass) do
    Class.new do
      def a
        is_a?(self)
      end
    end
  end

  before { allow_any_instance_of(klass).to receive(:is_a?).and_return(true) }

  it { expect(klass.new.a).to be(true) }
end
```

### Expected behavior

The spec should pass.

### Actual behavior

```shell
% rspec stub_is_a_spec.rb
Failures:

  1) AccountsProduction::AppSwitcherRequester 
     Failure/Error: is_a?(self)
     
     SystemStackError:
       stack level too deep
```

The stack is recursively called at this line of code obviously.
https://github.com/rspec/rspec-mocks/blob/v3.12.6/lib/rspec/mocks/proxy.rb#L38

I don't encounter the error if I initialize the class and utilize the instance generated.

undees:
Fixes #73.

The `Proxy` class in `rspec-mocks` uses `is_a?(Symbol)` to determine whether we can safely stub methods on a given receiver. This can cause an infinite loop when a class explicitly overrides `#is_a?`.

Now, when a class stubs `#is_a?`, we'll use the original, un-stubbed version of `#is_a?` to perform this eligibility check.

### DIFF
--- a/rspec-mocks/lib/rspec/mocks/proxy.rb
+++ b/rspec-mocks/lib/rspec/mocks/proxy.rb
@@ -32,7 +32,7 @@ module RSpec
 
       # @private
       def ensure_can_be_proxied!(object)
-        return unless object.is_a?(Symbol)
+        return unless Symbol === object
 
         msg = "Cannot proxy frozen objects. Symbols such as #{object} cannot be mocked or stubbed."
         raise ArgumentError, msg
--- a/rspec-mocks/spec/rspec/mocks/matchers/receive_spec.rb
+++ b/rspec-mocks/spec/rspec/mocks/matchers/receive_spec.rb
@@ -360,6 +360,56 @@ module RSpec
         # rubocop:enable Lint/StructNewOverride
       end
 
+      shared_examples "handles stubbed #is_a? cleanly for a single instance" do
+        let(:klass) do
+          Class.new do
+            def call_stubbed_method
+              is_a?(self.class)
+            end
+          end
+        end
+
+        let(:object) { klass.new }
+
+        it "safely checks that the instance can be proxied" do
+          stubbed_answer = double("stubbed on an instance")
+          target.to receive(:is_a?).and_return(stubbed_answer)
+          expect(object.call_stubbed_method).to eq(stubbed_answer)
+        end
+      end
+
+      shared_examples "handles stubbed #is_a? cleanly for any_instance_of" do
+        let(:superclass) { Class.new }
+
+        let(:klass) do
+          Class.new(superclass) do
+            def call_stubbed_method
+              is_a?(self.class)
+            end
+          end
+        end
+
+        context "the class under test" do
+          let(:stubbed_class) { klass }
+
+          it "safely checks that the instance can be proxied" do
+            stubbed_answer = double("stubbed on the class under test")
+            target.to receive(:is_a?).and_return(stubbed_answer)
+            expect(klass.new.call_stubbed_method).to eq(stubbed_answer)
+          end
+        end
+
+        context "the parent of the class under test" do
+          let(:stubbed_class) { superclass }
+
+          it "safely checks that the instance can be proxied" do
+            stubbed_answer = double("stubbed on the superclass")
+            target.to receive(:is_a?).and_return(stubbed_answer)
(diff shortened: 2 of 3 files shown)


rspec-mocks/lib/rspec/mocks/proxy.rb
```diff
@@ -31,13 +31,24 @@ def initialize(object, order_group, options={})
       end
 
       # @private
-      def ensure_can_be_proxied!(object)
-        return unless object.is_a?(Symbol)
+      def ensure_object_can_be_proxied!
+        return unless object_is_a_symbol?
```
JonRowe: I don't know why this wasn't done at the time, but the traditional RSpec solution to this problem is to use `Class ===` because its much less likely to be overridden:
```suggestion
        return unless Symbol === object
```

rspec-mocks/spec/rspec/mocks/matchers/receive_spec.rb
```diff
@@ -602,6 +602,49 @@ def receiver.method_missing(*); end # a poor man's stub...
...
+          let(:stubbed_answer) { double("stubbed on an instance")}
+          let(:instance) { klass.new }
+
+          before { allow(instance).to receive(:is_a?).and_return(stubbed_answer) }
+
+          it "uses the original #is_a? to check that the instance can be proxied" do
+            expect(instance.call_stubbed_method).to eq(stubbed_answer)
+          end
+        end
+      end
```
JonRowe: Wrap these into one of the existing contexts, they're regession checks for the various expectation types not a context on their own. I'd prefer one test for each type in its proper context.

JonRowe (changes requested):
:wave: Thanks for tackling this, I think theres a much more simpler way to solve this, see the suggested change for the guard condition.

If that works can you revert the other changes?

rspec-mocks/lib/rspec/mocks/proxy.rb
undees: Hi, I thought about using `===` but was concerned about someone mocking `Class#===` (though I agree with you that this is a rare case).

I don't mind switching it to `===`, but wanted to double-check with you about the tradeoffs. Are we okay with accepting the slight risk that someone may stub or override `Class.#===` in order to get the benefit of cleaner / more idiomatic Ruby?

pirj: If we're overly concerned, we can stash `===` and use the stashed one - we do this in a couple of places if I remeber correctly.

undees: I looked into the code, and there are other places (e.g., in `space.rb`) where we've got similar `Symbol#===` calls. These would require more time and risk to root out, so I think I'm okay with just doing `Symbol === object` in this PR as you suggest above.

22077c89fd8a67c3e0c23b19cf0cd31cd0bcdc55

rspec-mocks/spec/rspec/mocks/matchers/receive_spec.rb
undees: Good call, thank you! I was on the fence about whether these should be in a shared context, so your comment confirms it for me.

b65d001696196269421916bb635755797eb2d34b

undees:
> 👋 Thanks for tackling this, I think theres a much more simpler way to solve this, see the suggested change for the guard condition.
> 
> If that works can you revert the other changes?

Yup, happy to use a simpler approach per the discussion here. I think everything's in place for a second review.

JonRowe (approved):

JonRowe:
🎉 thanks!

JonRowe:
Released in 3.13.6

PR #270, opened 2025-10-20
Fix source_code_uri for the `rspec` gem

bquorning:
It seems “rspec-metagem” does not exist anymore, but which of these should the `source_code_uri` link to?

- https://github.com/rspec/rspec/blob/rspec-v3.13.1/rspec
- https://github.com/rspec/rspec


JonRowe:
Good catch thanks! (For consistency linking to the version like the others do seems practical, especially as its pointing to the gem)

PR #273, opened 2025-10-21
Fix predicate function scripts

bquorning:
`ruby -e "RUBY_ENGINE == 'ruby'"` will never return false, you need to wrap in `exit` to get the proper exit code.

~And when we already have the correct exit code, there is no need to wrap the condition with `if` and `return`.~

### DIFF
--- a/script/predicate_functions.sh
+++ b/script/predicate_functions.sh
@@ -1,10 +1,10 @@
 function is_mri {
-  if ruby -e "RUBY_ENGINE == 'ruby'"; then
+  if ruby -e "exit(RUBY_ENGINE == 'ruby')"; then
     # RUBY_ENGINE only returns 'ruby' on MRI.
     return 0
   else
     return 1
-  fi;
+  fi
 }
 
 function is_ruby_31_plus {


JonRowe:
I'm on the fence about this, the first might not look like it works but it does, and the second I'd rather the explicit statements.

```
ruby -v
# jruby 10.0.2.0 (3.4.2) 2025-08-07 cba6031bd0 OpenJDK 64-Bit Server VM 25 on 25 +indy +jit [arm64-darwin]
is_mri && echo "Is MRI" || echo "Is JRuby"
# Is JRuby
```

```
ruby -v
# ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM [arm64-darwin24]
is_mri && echo "Is MRI" || echo "Is JRuby"
# Is MRI
```

bquorning:
> the second I'd rather the explicit statements.

That’s fair, I’ll change that back.

> the first might not look like it works but it does, 

That is strange. I get a different result:
```
❯ docker run -ti --rm ruby:3.4 ruby -ve 'RUBY_ENGINE == "ruby"' && echo 'Is MRI' || echo 'Is JRuby'
ruby 3.4.7 (2025-10-08 revision 7a5688e2a2) +PRISM [aarch64-linux]
-e:1: warning: possibly useless use of == in void context
Is MRI
❯ docker run -ti --rm jruby:10.0 ruby -ve 'RUBY_ENGINE == "ruby"' && echo 'Is MRI' || echo 'Is JRuby'
jruby 10.0.2.0 (3.4.2) 2025-08-07 cba6031bd0 OpenJDK 64-Bit Server VM 21.0.8+9-LTS on 21.0.8+9-LTS +indy +jit [aarch64-linux]
-e:1: warning: possibly useless use of == in void context
Is MRI
```

JonRowe:
🤦 I just checked and on `main` you are correct, it is broken, I was running against 3-99-maintenance which does call exit

PR #275, opened 2025-10-30
Special case `to_h`/`to_hash` for null objects similar to `to_a`/`to_ary`

JonRowe:
Rails is now using `to_hash.as_json` for `Object#as_json` which causes an infinite loop for our spies which don't implement either `to_hash` or `as_json` themselves (they return spy which uses the Object implementation).

As we special case other such methods for null objects, we should do so for `to_hash` too.


JonRowe:
Released as rspec-mocks 3.13.7

PR #284, opened 2025-12-03
Pin bundler to 2.x

JonRowe:
Bundler 4.0.0 is not currently supported by aruba and is now general release for all Rubies > 3.2


pirj:
https://github.com/cucumber/aruba/pull/962 It does, just not released as a version yet

PR #285, opened 2025-12-08
Avoid building full backtrace when we only care about the direct caller

byroot:
`caller` without arguments build the full bakctrace, so it's performance is `O(n)` based on how deep the call stack is, and it can get very expensive.

Profiling of our app shows our test suite spends a bit of 1% of its runtime there.


JonRowe:
I'm wondering if this is worth pulling into our support library to make it easier to document this is for performance reasons, or otherwise add a benchmark, I'm happy to merge this without those changes if you'd rather not work on that howerver

byroot:
I mean, it's a very well known pattern. So yeah, I don't think it's worth adding benchmarks.

Issue #290, opened 2025-12-31
diff-lcs 2.0.0.beta.1 released


halostatue:
I've released diff-lcs 2.0.0.beta.1 today, with stable 2.0.0 planned for 2026-01-31.

The main breaking changes are minimum Ruby 3.2 and removal of `Diff::LCS.LCS` (use `Diff::LCS.lcs` instead). The primary API is otherwise unchanged.

Please test against the beta if you can. RSpec hasn't been updated to support the beta yet, so there's a circular dependency issue for now.

If you want to support both versions, you'll need `>= 1.6.0, < 3` in your dependency constraint.

See [PR #214](https://github.com/halostatue/diff-lcs/pull/214) for full details.

JonRowe:
👋 is there a specific reason / feature you're pinning to Ruby 3.2 (other than being EOL)? Historically supporting multiple versions of diff-lcs has been quite painful for us...

halostatue:
Data classes, plus EOL and a chance to modernize the code base (including with some pattern matching).

Many of the underlying objects created during comparison are never modified and would benefit from an immutable representation.

JonRowe:
Fair enough, I'll have to think about our approach, it might be we pin to 1.x in RSpec 4 and wait until a future version to raise our own Ruby version requirement to allow 2.x or extract support entirely to allow RSpec to be agnostic to the version/diff library

halostatue:
FWIW, I think that the main API is completely compatible between both versions of diff-lcs.

If it isn't, please file a bug—my goal is to make sure that rspec on modern ruby can take advantage of the modern implementation but that the upgrade should otherwise be invisible.

(There are other things removed, but as one of my contributors said last year, they've been broken for a while with no bug reports on them.)

JonRowe:
It's not just the API its the output, changes to the output heavily affect our specs because its asserted on, in 3.x we have a lot of helpers to cope with the differences in output across the various versions, some forced by us testing against the different versions to ensure it works some forced by ruby versions. Given I don't know the timespan between 4.0 and 4.1 when we might drop Ruby up to 3.2 I don't want to have to maintain that sort of fix again, I'd rather refactor the whole thing to be less dependent on the differ, but I don't want that to hold up RSpec 4.

halostatue:
Fair enough, but I would consider the diff output to be part of the API and changes here would be breaking changes.

The only things removed are:

- htmldiff (subtly broken and there are better solutions)
- ed and reverse ed diff output

[...]

PR #291, opened 2026-01-01
Allow for diff-lcs 2.x

pirj:
See #290


mvz:
@pirj thanks for working on this. Please note that Cucumber still depends on diff-lcs ~> 1.5, so will force the 1.6 version to be installed.

PR #297, opened 2026-01-23
[rspec-mocks] prevent `any_instance` from stubbing private/protected methods defined in prepended modules

### LINKED ISSUE #293: `allow_any_instance` + prepended private method leads to state leakage between tests
## Problem

rspec-mocks attempts to prevent usage of `allow_any_instance` on methods that have been defined in a prepended module.

  * detected here: https://github.com/rspec/rspec-mocks/blob/3caa9ac841e146d618421f870077c0ad684e6cad/lib/rspec/mocks/any_instance/recorder.rb#L285-L289
  * reported here: https://github.com/rspec/rspec-mocks/blob/3caa9ac841e146d618421f870077c0ad684e6cad/lib/rspec/mocks/any_instance/error_generator.rb#L20-L22

This check is not successful for **private methods** which have been defined in prepended module, because the check is based only on `method_defined?`, which does not detect private methods.

```ruby
class TestSubject
  def sample_public_method; end

  private

  def sample_private_method; end
end

TestSubject.method_defined?(:sample_public_method)
# => true

TestSubject.method_defined?(:sample_private_method)
# => false
```

## Impact

`allow_any_instance` stubs which are set up for prepended private methods are not correctly cleaned up. the `allow_any_instance` stub remains in place in later tests.

this caused a flakey failure for me when one of these stubbed methods was invoked in a `before(:context)` block, by the [`let_it_be` test helper](https://test-prof.evilmartians.io/recipes/let_it_be). 

since this is outside any example, and rspec is in `RootSpace`, the resulting error was

```
"undefined method 'ensure_registered' for RSpec::Mocks::RootSpace"
```

since this only surfaces when tests are run in a specific order, the behavior is flakey.

## Reproduction

The test case below shows how to reproduce the issue.

Disclosure: I did use Cursor to help explore the issue and draft this reproduction case. I've iterated on it a bunch to try to make the reproduction as clear as possible, but as stated above there's a lot I'm not sure about when it comes to rspec internals and I might have some points mixed up here.

```
ruby rspec_mocks_prepend_bug_spec.rb
```

For ease of reproduction and to minimize dependencies, I've omitted `let_it_be` from this script. Functionally, this should be the same as when `let_it_be` is used.

```ruby
# frozen_string_literal: true

# Self-contained reproduction of rspec-mocks bug with any_instance + prepended private methods
#
# Run with:
#   ruby rspec_mocks_prepend_bug_spec.rb
#
# THE BUG:
# When a class has a prepended module that defines a PRIVATE method, rspec-mocks'
# any_instance check (`allow_no_prepended_module_definition_of`) fails to detect it
# because it only uses `method_defined?` which returns false for private methods.
#
# This allows the any_instance stub to be set up, but:
# 1. The stub doesn't actually work (the prepended module shadows it)
# 2. More critically: cleanup doesn't work correctly, leaving the class in a broken
#    state where the any_instance-defined method persists and tries to call into the
#    mock system even when it's inactive (RootSpace)
#
# IMPACT:
# When combined with `before(:context)` (which runs outside the mock lifecycle),
# this causes: "undefined method 'ensure_registered' for RSpec::Mocks::RootSpace"
#
# the bug requires:
#   1. A prepended module defining a private method
#   2. Using any_instance_of to stub that same method
#   3. Later running code in before(:context) that triggers the method

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  gem 'rspec', '~> 3.13'
end

require 'rspec/autorun'

RSpec.configure do |config|
  config.order = :defined
  config.formatter = :documentation
end

#------------------------------------------------------------------------------
# Setup: A module that wraps methods (similar to CallbackTracer in the real codebase)
#------------------------------------------------------------------------------

module InstrumentationWrapper
  def self.wrap_method!(klass, method_name)
    visibility = if klass.private_method_defined?(method_name)
                   :private
                 elsif klass.protected_method_defined?(method_name)
                   :protected
                 else
                   :public
                 end

    wrapper_module = Module.new do
      define_method(method_name) do |*args, **kwargs, &block|
        @instrumentation_called = true
        super(*args, **kwargs, &block)
      end
      # give it a name to make inspection/debugging clearer - so we don't just see `#<Module:0x...>`
      define_singleton_method(:inspect) { "InstrumentationWrapper for #{klass}##{method_name}" }
      send(visibility, method_name)
    end

    klass.prepend(wrapper_module)
  end
end

class TestSubject
  attr_reader :result, :instrumentation_called

  def initialize
    @result = nil
    @instrumentation_called = false
  end

  def do_work
    perform_internal_operation
    @result
  end

  private

  def perform_internal_operation
    @result = "original_result"
  end
end

# Wrap the PRIVATE method after it's defined
InstrumentationWrapper.wrap_method!(TestSubject, :perform_internal_operation)

# A class WITHOUT any prepended module (for comparison)
class TestSubjectWithoutPrepend
  attr_reader :result

  def initialize
    @result = nil
  end

  def do_work
    perform_internal_operation
    @result
  end

  private

  def perform_internal_operation
    @result = "original_result"
  end
end

#------------------------------------------------------------------------------
# Test 1: Demonstrate that rspec-mocks correctly catches PUBLIC prepended methods
#------------------------------------------------------------------------------

RSpec.describe "BASELINE: rspec-mocks catches PUBLIC prepended methods" do
  module PublicWrapper
    def public_method
      super
    end
  end

  class TestWithPublicPrepend
    def public_method
      "original"
    end
    prepend PublicWrapper
  end

  it "raises an error when trying to stub a public method on a prepended module" do
    # expected error: "Using `any_instance` to stub a method (public_method) that has been defined on a prepended module (PublicWrapper) is not supported."
    expect {
      allow_any_instance_of(TestWithPublicPrepend).to receive(:public_method)
    }.to raise_error(RSpec::Mocks::MockExpectationError, /prepended module/)
  end
end

#------------------------------------------------------------------------------
# Test 2: Demonstrate the bug - PRIVATE prepended methods are NOT caught
#------------------------------------------------------------------------------

RSpec.describe "BUG: rspec-mocks MISSES private prepended methods" do
  it "incorrectly allows stubbing a private method defined in a prepended module" do
    # This SHOULD raise an error like the public case above, but it doesn't.
    expect {
      allow_any_instance_of(TestSubject).to receive(:perform_internal_operation)
    }.to raise_error(RSpec::Mocks::MockExpectationError)
  end
end

#------------------------------------------------------------------------------
# Test 3: Demonstrate the CONSEQUENCE - broken state with before(:context)
# Run with --order defined to ensure correct ordering
#------------------------------------------------------------------------------

RSpec.describe "FULL RECREATION SCENARIO: broken mock state persists to before(:context)" do
  context "first: sets up any_instance stub on private prepended method" do
    before do
      # ┌─────────────────────────────────────────────────────────────────────────┐
      # │ DEBUGGER STAGE 1: Before the stub - clean state                         │
      # │                                                                         │
      # │ Run these commands:                                                     │
      # │   TestSubject.instance_method(:perform_internal_operation).owner        │
      # │   # => InstrumentationWrapper for TestSubject#perform_internal_operation│
      # │   #    (the prepended wrapper module owns the method)                   │
      # │                                                                         │
      # │   TestSubject.instance_method(:perform_internal_operation).source_location
      # │   # => ["...rspec_mocks_prepend_bug_spec.rb", 46]                        │
      # │   #    (points to the wrapper's define_method)                          │
      # │                                                                         │
      # │   RSpec::Mocks.space.class                                              │
      # │   # => RSpec::Mocks::Space  (active mock space - stubs allowed)         │
      # │                                                                         │
      # │   TestSubject.new.tap(&:do_work).result                                 │
      # │   # => "original_result"  (method works fine)                           │
      # └─────────────────────────────────────────────────────────────────────────┘
      # debugger # STAGE 1: Before stub

      # This is allowed (bug!) but leaves the class in a broken state after cleanup
      allow_any_instance_of(TestSubject).to receive(:perform_internal_operation)

      # ┌─────────────────────────────────────────────────────────────────────────┐
      # │ DEBUGGER STAGE 2: After the stub - see the broken recorder              │
      # │                                                                         │
      # │ Run these commands:                                                     │
      # │   recorder = RSpec::Mocks.space.instance_variable_get(:@any_instance_recorders).values.last
      # │                                                                         │
      # │   recorder.instance_variable_get(:@klass)                               │
      # │   # => InstrumentationWrapper for TestSubject#perform_internal_operation│
      # │   #    BUG! This should be TestSubject, not the wrapper module!         │
      # │                                                                         │
      # │   recorder.instance_variable_get(:@klass) == TestSubject                │
      # │   # => false  (confirms @klass is wrong)                                │
      # │                                                                         │
      # │   recorder.instance_variable_get(:@backed_up_method_owner)              │
      # │   # => {}  (empty! nothing was backed up because of the confusion)      │
      # │                                                                         │
      # │   recorder.instance_variable_get(:@observed_methods)                    │
      # │   # => []  (empty! method wasn't properly observed)                     │
      # └─────────────────────────────────────────────────────────────────────────┘
      # debugger # STAGE 2: After stub - broken recorder state
    end

    it "runs and leaves broken state behind" do
      # The test itself runs fine
      TestSubject.new.do_work
    end
  end

  context "second: uses before(:context) which runs outside mock lifecycle" do
    # before(:context) runs when RSpec::Mocks.space is RootSpace.
    # The broken any_instance machinery from the previous test tries to call
    # into the mock system, causing the error.
    #
    # Expected error:
    #   NoMethodError: undefined method 'ensure_registered' for RSpec::Mocks::RootSpace
    before(:context) do
      # ┌─────────────────────────────────────────────────────────────────────────┐
      # │ DEBUGGER STAGE 3: In before(:context) - RootSpace, method call will fail│
      # │                                                                         │
      # │ Run these commands:                                                     │
      # │   RSpec::Mocks.space.class                                              │
      # │   # => RSpec::Mocks::RootSpace  (inactive - no stubs allowed here)      │
      # │                                                                         │
      # │   RSpec::Mocks.space.instance_variable_get(:@any_instance_recorders)    │
      # │   # => nil  (recorders were "cleaned up" but left broken state behind)  │
      # │                                                                         │
      # │   TestSubject.private_instance_methods(false).grep(/perform/)           │
      # │   # => [:__perform_internal_operation_without_any_instance__,           │
      # │   #     :perform_internal_operation]                                    │
      # │   # The backup method shouldn't exist after cleanup! Evidence of bug.   │
      # │                                                                         │
      # │   TestSubject.instance_method(:__perform_internal_operation_without_any_instance__).source_location
      # │   # => ["...spec.rb", 74]  (the ORIGINAL method, renamed by rspec-mocks)│
      # │                                                                         │
      # │   TestSubject.new.send(:perform_internal_operation)                     │
      # │   # => error: undefined method 'ensure_registered' for an instance of   │
      # │   #    RSpec::Mocks::RootSpace                                          │
      # │   #    rspec/mocks/any_instance/recorder.rb:110:in 'Recorder#playback!' │
      # │   #    rspec/mocks/any_instance/recorder.rb:264:in 'TestSubject#observe!'
      # │   #    rspec_mocks_prepend_bug_spec.rb:48:in 'wrap_method!'             │
      # │                                                                         │
      # │   # The hook on TestSubject (installed by observe!) was never cleaned   │
      # │   # up, so when wrapper calls super(), it triggers rspec-mocks code     │
      # │   # that tries to use the inactive RootSpace.                           │
      # └─────────────────────────────────────────────────────────────────────────┘
      # debugger # STAGE 3: In before(:context) - about to trigger the error
      @subject_instance = TestSubject.new.tap(&:do_work)
    end

    it "FAILS with RootSpace error due to lingering any_instance state" do
      expect(@subject_instance.result).to eq("original_result")
    end
  end
end

#------------------------------------------------------------------------------
# Test 4: Demonstrate that WITHOUT prepend, any_instance on private methods works fine
#------------------------------------------------------------------------------

RSpec.describe "CONTROL: without prepend, any_instance on private methods works correctly" do
  context "first: sets up any_instance stub on private method (no prepend)" do
    before do
      allow_any_instance_of(TestSubjectWithoutPrepend).to receive(:perform_internal_operation)
    end

    it "works fine during the test" do
      TestSubjectWithoutPrepend.new.do_work
    end
  end

  context "second: uses before(:context) - works correctly because no prepend interference" do
    # Without a prepend, the any_instance cleanup works correctly,
    # so before(:context) has no issues.
    before(:context) do
      @subject_instance = TestSubjectWithoutPrepend.new.tap(&:do_work)
    end

    it "works fine - no RootSpace error" do
      expect(@subject_instance.result).to eq("original_result")
    end
  end
end
```

alexdean:
fixes rspec/rspec#293

### DIFF
--- a/rspec-mocks/lib/rspec/mocks/any_instance/recorder.rb
+++ b/rspec-mocks/lib/rspec/mocks/any_instance/recorder.rb
@@ -276,7 +276,9 @@ module RSpec
 
         def allow_no_prepended_module_definition_of(method_name)
           prepended_modules = RSpec::Mocks::Proxy.prepended_modules_of(@klass)
-          problem_mod = prepended_modules.find { |mod| mod.method_defined?(method_name) }
+          problem_mod = prepended_modules.find do |mod|
+            MethodReference.method_defined_at_any_visibility?(mod, method_name)
+          end
           return unless problem_mod
 
           AnyInstance.error_generator.raise_not_supported_with_prepend_error(method_name, problem_mod)
--- a/rspec-mocks/spec/rspec/mocks/any_instance_spec.rb
+++ b/rspec-mocks/spec/rspec/mocks/any_instance_spec.rb
@@ -178,34 +178,92 @@ module RSpec
         end
 
         context "when the class has a prepended module" do
-          it 'allows stubbing a method that is not defined on the prepended module' do
-            klass.class_eval { prepend Module.new { def other; end } }
-            allow_any_instance_of(klass).to receive(:foo).and_return(45)
+          context 'with a method not defined on the prepended module' do
+            before do
+              klass.class_eval { prepend Module.new { def other; end } }
+            end
 
-            expect(klass.new.foo).to eq(45)
-          end
-
-          it 'prevents stubbing a method that is defined on the prepended module' do
-            klass.class_eval { prepend Module.new { def foo; end } }
-
-            expect {
+            it 'allows stubbing the method' do
               allow_any_instance_of(klass).to receive(:foo).and_return(45)
-            }.to fail_with(/prepended module/)
-          end
 
-          it 'allows stubbing a chain starting with a method that is not defined on the prepended module' do
-            klass.class_eval { prepend Module.new { def other; end } }
-            allow_any_instance_of(klass).to receive_message_chain(:foo, :bar).and_return(45)
+              expect(klass.new.foo).to eq(45)
+            end
 
-            expect(klass.new.foo.bar).to eq(45)
-          end
-
-          it 'prevents stubbing a chain starting with a method that is defined on the prepended module' do
-            klass.class_eval { prepend Module.new { def foo; end } }
-
-            expect {
+            it 'allows stubbing a chain starting with the method' do
               allow_any_instance_of(klass).to receive_message_chain(:foo, :bar).and_return(45)
-            }.to fail_with(/prepended module/)
+
+              expect(klass.new.foo.bar).to eq(45)
+            end
+          end
+
+          context 'with a public method defined on the prepended module' do
+            before do
+              klass.class_eval { prepend Module.new { def foo; end } }
+            end
+
+            it 'prevents stubbing the method' do
+              expect {
+                allow_any_instance_of(klass).to receive(:foo).and_return(45)
+              }.to fail_with(/prepended module/)
+            end
+
+            it 'prevents stubbing a chain starting with the method' do
+              expect {
+                allow_any_instance_of(klass).to receive_message_chain(:foo, :bar).and_return(45)
+              }.to fail_with(/prepended module/)
+            end
+          end
+
+          context 'with a protected method defined on the prepended module' do
+            before do
+              klass.class_eval do
+                prepend Module.new {
+                  protected
+
+                  def protected_method
+                    :prepended_protected_method_return_value
+                  end
+                }
+              end
+            end
+
+            it 'prevents stubbing the method' do
+              expect {
+                allow_any_instance_of(klass).to receive(:protected_method).and_return(45)
+              }.to fail_with(/prepended module/)
+            end
+
+            it 'prevents stubbing a chain starting with the method' do
+              expect {
+                allow_any_instance_of(klass).to receive_message_chain(:protected_method, :bar).and_return(45)
+              }.to fail_with(/prepended module/)
+            end
+          end
+
+          context 'with a private method defined on the prepended module' do
+            before do
+              klass.class_eval do
+                prepend Module.new {
+                  private
+
+                  def private_method
+                    :prepended_private_method_return_value
+                  end
+                }
+              end
+            end
+
+            it 'prevents stubbing the method' do
+              expect {
+                allow_any_instance_of(klass).to receive(:private_method).and_return(45)
+              }.to fail_with(/prepended module/)
+            end
+
+            it 'prevents stubbing a chain starting with the method' do
+              expect {
+                allow_any_instance_of(klass).to receive_message_chain(:private_method, :bar).and_return(45)
+              }.to fail_with(/prepended module/)
+            end
           end
         end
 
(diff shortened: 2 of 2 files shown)


corsonknowles:
Thanks for finding and patching this

alexdean:
@JonRowe hello sorry to bother but wondering if you might be able to review/approve this? (or provide guidance on what needs to be changed.) thank you!

alexdean:
@JonRowe At Gusto, we've run this change as a monkeypatch in a CI branch 10 times over the past month (by merging `main` into the branch periodically) and have not seen any surprising behavior. 

We're waiting to merge our patch pending some feedback on this PR, as we don't want to diverge from mainline rspec behavior.

JonRowe:
Apologies for the delay here, it got lost somewhere along the line and I've been meaning to get to it all week since your ping.

JonRowe:
Released in 3.13.8

corsonknowles:
Really appreciate you Jon! Thanks for all the help here. 

We'll likely lean hard into using let_it_be in certain contexts after this. 

Will let you know if we find any other unusual interactions, but hopefully it's pure smooth sailing from here.

PR #298, opened 2026-01-24
FuzzyMatcher compares to a Range via ==

### LINKED ISSUE #233: Range comparison using `have_attributes` apparently broken
When I test objects that has an attribute which is an array of ranges, the `have_attributes` matcher doesn't detect difference when expected range contains the actual range. So this fails as it should:
```ruby
let(:has_ranges) { OpenStruct.new(ranges: [0...10]) }

it 'detects incorrect range' do
  expect(has_ranges).to have_attributes(ranges: [1...10])
end
```

But this passes fine:
```ruby
let(:has_ranges) { OpenStruct.new(ranges: [0...10]) }

it 'detects incorrect range' do
  expect(has_ranges).to have_attributes(ranges: [-1...11])
end
```

The problem seems to be in this file:
rspec/support/fuzzy_matcher.rb
```
10|       def self.values_match?(expected, actual)
11|         if Hash === actual
12|           return hashes_match?(expected, actual) if Hash === expected
13|         elsif Array === expected && Enumerable === actual && !(Struct === actual)
14|           return arrays_match?(expected, actual.to_a)
15|         end
16| 
17|         return true if expected == actual
18| 
19|         begin
20|           expected === actual
21|         rescue ArgumentError
22|           # Some objects, like 0-arg lambdas on 1.9+, raise
23|           # ArgumentError for `expected === actual`.
24|           false
25|         end
26|       end
```

So ranges are compared at line 20 using `===` which tests whether lhs covers rhs

My setup:
```
rspec-core (3.13.5)
      rspec-support (~> 3.13.0)
    rspec-expectations (3.13.3)
      rspec-support (~> 3.13.0)
    rspec-mocks (3.13.2)
      rspec-support (~> 3.13.0)
    rspec-rails (6.0.4)
      rspec-core (~> 3.12)
      rspec-expectations (~> 3.12)
      rspec-mocks (~> 3.12)
      rspec-support (~> 3.12)
    rspec-support (3.13.4)
    rubocop-rspec (3.3.0)
  rspec-rails (~> 6.0.0.rc1)
```

nevinera:
The default behavior of using `expected === actual` doesn't behave as one would expect for the case where your expected value is an instance of Range, which causes everything _relying_ on FuzzyMatcher to give odd results for that case.

The prompting example is:

    expect(OpenStruct.new(a: (1..5))).to have_attributes(a: (1..3))

which passes (since `(1..5) === (1..3)`). Range has a specific implementation of triple-equals, owing to its use in case-statements (eg `when (1..9) then :small`), but the semantics of that implementation do not match what one typically needs when expressing a value-matcher in a test.

Note that the `be_between` and `cover` composable matchers already exist, so while this _will_ cause this to break:

    expect(OpenStruct.new(a: 5)).to have_attributes(a: (1..100))

that is already better expressed in another way.

Fixes #233

### DIFF
--- a/rspec-support/lib/rspec/support/fuzzy_matcher.rb
+++ b/rspec-support/lib/rspec/support/fuzzy_matcher.rb
@@ -10,6 +10,9 @@ module RSpec
       def self.values_match?(expected, actual)
         if Hash === actual
           return hashes_match?(expected, actual) if Hash === expected
+        elsif Range === expected
+          # Check that actual is an equivalent Range, rather than included/covered.
+          return expected == actual
         elsif Array === expected && Enumerable === actual && !(Struct === actual)
           return arrays_match?(expected, actual.to_a)
         end
--- a/rspec-support/spec/rspec/support/fuzzy_matcher_spec.rb
+++ b/rspec-support/spec/rspec/support/fuzzy_matcher_spec.rb
@@ -27,6 +27,13 @@ module RSpec
         expect(/foo/).not_to match_against(/bar/)
       end
 
+      it 'can match a range only against a range' do
+        expect(1..5).to match_against(1..5)
+        expect(1..5).not_to match_against(1..6)
+        expect(1..5).not_to match_against(2..4)
+        expect(1..5).not_to match_against(3)
+      end
+
       it 'can match a class against an instance' do
         expect(String).to match_against("foo")
         expect(String).not_to match_against(123)


JonRowe:
Thanks 🎉

JonRowe:
Released in rspec-support 3.13.7

sfcgeorge:
This caused one of our specs to fail using the "change by" matcher:

```ruby
     Failure/Error:
       expect do
         subject
       end.to change { TimeSlotGroup.count }.by(11..12)
     
       expected `TimeSlotGroup.count` to have changed by 11..12, but was changed by 12
```

I thought maybe I could do this instead but it's not a chainable matcher

```ruby
  .to change { TimeSlotGroup.count }.by_at_least(11).by_at_most(12) # nope
```

However I found several ways to achieve the same result by composing matchers:

```ruby
.to change { TimeSlotGroup.count }.by(an_object_satisfying { (11..12).cover?(it) })
.to change { TimeSlotGroup.count }.by(a_value_between(11, 12)) # nice and clear
```

**Thoughts**

* I think this is a good change overall 👍 
* I don't know if `by` was ever supposed to work the way I was using it, the docs don't show that use-case.
* I'm happy with the `a_value_between` option, nothing necessarily needs to change here I'm just noting it.
* Some additional `by` variations could be added, though these read a bit tortured (and composing matchers already works anyway):
  * change().by_between(minimum, maximum).inclusive
  * change().by_within(delta).of(expected)
* This possibly shouldn't have been a patch release as it's a breaking change but I'm not suggesting you pull it or anything.
* It was surprising that updating a sub-dependency of RSpec caused RSpec test failures - perhaps bumping the RSpec version in lockstep would be clearer albeit ceremony.
* Yes we should fix our code to always generate a full year's worth of things and not sometimes 11.

JonRowe:
Thanks for adding this to the collective knowledge, I'll have a think about wether `by` should actually support this officially, it could easily short circuit the fuzzy check.

> This possibly shouldn't have been a patch release as it's a breaking change but I'm not suggesting you pull it or anything.

I disagree with this definition, if something is unspecified behaviour its not a breaking change if it changes.

> It was surprising that updating a sub-dependency of RSpec caused RSpec test failures - perhaps bumping the RSpec version in lockstep would be clearer albeit ceremony.

I'm sorry this was surprising but for what its worth there is currently no single "RSpec version" to bump in lockstep, it is something I'm considering for 4.x for change management reasons as well as the principle of least surprise.

PR #303, opened 2026-02-05
Use Prism for parsing

kddnewton:
There are two places where RSpec uses Ripper for parsing Ruby source: SnippetExtractor and BlockSnippetExtractor. This commit changes them to use Prism instead. A couple of benefits right off the bat:

* The implementation is much simpler. It is about 20 LOC for SnippetExtractor with Prism, and 30 LOC for BlockSnippetExtractor. For Ripper it is about 100 LOC for SnippetExtractor + management of tokens and nodes. For BlockSnippetExtract it is about 200 LOC
  + management of a parsing state machine.
* It is more correct. There is a test failure where it knows it cannot parse the heredoc, but with Prism you get that for free. There is also a place where it was incorrectly attributing a block for the BlockSnippetExtractor in the test, because {} and do/end are not interchangeable in the context of a method call without parentheses. This all comes for free.
* It is faster, sometimes by orders of magnitude depending on the file being parsed.
* It is more portable. CRuby, JRuby, and TruffleRuby will all support this out of the box, so you no longer have to worry about detecting the implementation at runtime.

There is a question here of if you want to drop support for the Ripper implementation entirely (3.2 is EOL in a month). In this commit I have left it in place, prioritizing Prism if it is available.

There are two remaining test failures:
* I wasn't sure if you wanted to explicitly reject there heredoc case or not. Happy to go either way.
* I wasn't sure exactly how you handle requires and sandboxing. I'm sure there's a way to list Prism as an allowed require, but I can't figure out exactly where to put that.

### DIFF
--- a/rspec-core/lib/rspec/core/formatters/snippet_extractor.rb
+++ b/rspec-core/lib/rspec/core/formatters/snippet_extractor.rb
@@ -1,4 +1,5 @@
 RSpec::Support.require_rspec_support 'ruby_features'
+RSpec::Support.require_rspec_support 'source'
 
 module RSpec
   module Core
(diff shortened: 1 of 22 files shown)


rspec-core/lib/rspec/core/formatters/snippet_extractor.rb
```diff
@@ -20,7 +20,31 @@ def self.source_from_file(path)
           RSpec.world.source_from_file(path)
         end
 
-        if RSpec::Support::RubyFeatures.ripper_supported?
+        if RSpec::Support::RubyFeatures.prism_supported?
+          require "prism"
```
JonRowe: At the very least, this should be moved inside the method the same way ripper has its require, the reason for this is to delay the loading until such time where if code reasonably uses it, it has loaded it already. This avoids a false positive where we require things silently and the code appears to work but then crashes outside of the test environment.

The check for require is a shared example in rspec-support and if you grep for `allow_loaded_features` you'll find the other libraries that are preloaded, but ripper is actually absent because it is delayed loading as described and thus escapes this check.

JonRowe (commented):
Thanks for this! In general I'm happy about this change, we need to keep the ripper version for 3.0-3.2 support which we are not currently planning on dropping in RSpec 4 but may happen later.

That does lead to a slight concern about the `Source` class which will still try to load ripper if it is accessed...

kddnewton:
> Thanks for this! In general I'm happy about this change, we need to keep the ripper version for 3.0-3.2 support which we are not currently planning on dropping in RSpec 4 but may happen later.
> 
> That does lead to a slight concern about the `Source` class which will still try to load ripper if it is accessed...

Hey Jon — that's great news! I will try to implement the things you mentioned above, and I'll see if I can get the Source class a little under control. I was trying to leave things mostly in place to not rock the boat too much, but I agree you wouldn't want that to load ripper immediately if we're not actually going to use it.

kddnewton:
@JonRowe I've implemented your changes, and I think everything should be green now. Source now prioritizes prism as well, so we should be good on that front.

kddnewton:
* @JonRowe line coverage has dropped since we don't explicitly test the parser implementation — we just use the parser that is available. So in instances where Prism is available, we use it, which means the Ripper implementation is not used. I'm happy to make it test both parsers in all environments if that's preferred, or I can mark Ripper as not required line coverage. Your choice.
* @enebo it looks like JRuby 10 you can't require prism by default. Do you have to have it in the gemfile to be available? I can turn this off for JRuby but I figured I'd check.

JonRowe:
> I'm happy to make it test both parsers in all environments if that's preferred, or I can mark Ripper as not required line coverage. Your choice.

Given ripper is used in some environments, I would like both paths covered please.

kddnewton:
Hey @JonRowe — I've split things into different classes so that both prism and ripper can be tested in the test suite. We should be good to go now I think.

kddnewton:
Sorry @JonRowe one more time... this time I checked all the rubies and JRuby, hopefully this is the last one...

kddnewton:
@JonRowe sorry to be a pain here. It's working as expected now — both parsers are tested in all versions 3.3+. But because Prism isn't supported in 3.2 and below, the test coverage drops for those CI runs. Should I lower the required minimum coverage for those rubies? Or try another approach? Happy with whatever you like here, but unfortunately I think I need guidance from you on how to proceed.

kddnewton:
Nice that's great. I think this is good to go then unless you have any other feedback.

PR #305, opened 2026-02-11
Add end-of-run counts to summary line

### LINKED ISSUE #4: Show number of assertions in output.
I searched around, but couldn't find a way to do it. Something like:

```
1 example, 0 failures, 3 assertions
```

In [this question](http://stackoverflow.com/questions/6137874/show-number-of-assertions-in-rspec-output), @dchelimsky suggested to open a feature request, so here I am :smiley_cat: 

_Edit: @xaviershay's summary of comments below as of Jan 2017:_

* Proposed API: https://github.com/rspec/rspec/issues/4
* External gem that claims to do this (I haven't verified): https://github.com/bootstraponline/rspec_expectation_count

tangposmarvin:
Closes rspec/rspec#4                                                                                                                                                                        

## Summary                                                                                                                                                                                  
                                                                                                                                                                                            
Adds a public API for libraries to register custom counters that appear in the
end-of-run summary line alongside examples, failures, and pending.

rspec-expectations uses this to display expectation counts:

    3 examples, 0 failures, 5 expectations

## Changes

### rspec-core

- `Configuration#add_end_of_run_count(singular, plural=nil, &block)` — public
  API for registering counters
- `EndOfRunCount` value object with `to_s`, `as_json`, and
  `self.join_for_summary` for formatting
- `SummaryNotification` gains `end_of_run_counts` accessor, wired through
  `Reporter#finish`
- HTML, JSON, and text formatters all render the extra counts
- Graceful handling: blocks that raise get a warning and are skipped, blocks
  that return nil are omitted

### rspec-expectations

- `RSpec::Expectations.expectation_count` counter incremented in both
  `PositiveExpectationHandler` and `NegativeExpectationHandler`
- Registers via `add_end_of_run_count("expectation")` at require time
- Counter reset in `before(:suite)` to support multiple `Runner.run` calls

## Test plan

- Unit specs for `add_end_of_run_count` (singular, plural, custom plural,
  missing block, multiple counters)
- Unit specs for `SummaryNotification#totals_line` (no counts, single, multiple,
  zero, singular, custom plural, error, nil)
- Unit specs for `SummaryNotification#end_of_run_counts` nil-safe default
- Unit specs for JSON formatter extra counts in summary hash
- Unit spec for Reporter passing counts through to SummaryNotification
- Unit specs for expectation counting (positive, negative, accumulation, reset)
- Integration coverage via `clear_examples.feature` (counter resets across runs)
- Integration coverage via `json_formatter.feature` (summary_line includes count)

### DIFF
--- a/rspec-core/lib/rspec/core/configuration.rb
+++ b/rspec-core/lib/rspec/core/configuration.rb
@@ -6,6 +6,39 @@ RSpec::Support.require_rspec_support 'ruby_features'
 
 module RSpec
   module Core
+    # @private
+    # Value object representing a counter to display at end of run.
+    EndOfRunCount = Struct.new(:singular, :plural, :block) do
+      def self.join_for_summary(counts)
+        suffix = counts.map(&:to_s).reject(&:empty?).join(", ")
+        suffix.empty? ? "" : ", #{suffix}"
+      end
+
+      def to_s
+        return "" unless compute_count
+        "#{@count} #{label}"
+      end
+
+      def as_json
+        return {} unless compute_count
+        { :"#{singular.gsub(' ', '_')}_count" => @count }
+      end
+
+    private
+
+      def compute_count
+        @count = block.call
+        !@count.nil?
+      rescue => e
+        RSpec.warning("End-of-run counter #{singular.inspect} raised: #{e.message}")
+        false
+      end
+
+      def label
+        @count.to_i == 1 ? singular : (plural || "#{singular}s")
+      end
+    end
+
     # rubocop:disable Metrics/ClassLength
 
     # Stores runtime configuration information.
@@ -429,7 +462,8 @@ module RSpec
       # @private
       attr_accessor :static_config_filter_manager
       # @private
-      attr_reader :backtrace_formatter, :ordering_manager, :loaded_spec_files
+      attr_reader :backtrace_formatter, :ordering_manager, :loaded_spec_files,
+                  :end_of_run_counts
 
       # rubocop:disable Metrics/AbcSize
       # rubocop:disable Metrics/MethodLength
@@ -489,6 +523,7 @@ module RSpec
         @world = World::Null
         @pending_failure_output = :full
         @force_line_number_for_spec_rerun = false
+        @end_of_run_counts = []
         @warnings = nil
 
         if warnings?
@@ -580,6 +615,24 @@ module RSpec
           add_setting(name, opts)
         end
         __send__("#{name}=", default) if default
+      end
+
+      # Registers a counter to be displayed in the end-of-run summary line.
+      # The block should return the current count when called.
+      #
+      # @param singular [String] singular label (e.g. "expectation")
+      # @param plural [String] optional plural label; defaults to singular + "s"
+      # @yield block that returns the current count
+      #
+      # @example
+      #     RSpec.configure do |c|
+      #       c.add_end_of_run_count("expectation") do
+      #         MyLib.expectation_count
+      #       end
+      #     end
+      def add_end_of_run_count(singular, plural=nil, &block)
+        raise ArgumentError, "A block that returns the count is required" unless block
+        @end_of_run_counts << EndOfRunCount.new(singular, plural, block)
       end
 
       # Returns the configured mock framework adapter module.
--- a/rspec-core/lib/rspec/core/configuration.rb
+++ b/rspec-core/lib/rspec/core/configuration.rb
@@ -6,39 +6,6 @@ RSpec::Support.require_rspec_support 'ruby_features'
 
 module RSpec
   module Core
-    # @private
-    # Value object representing a counter to display at end of run.
-    EndOfRunCount = Struct.new(:singular, :plural, :block) do
-      def self.join_for_summary(counts)
-        suffix = counts.map(&:to_s).reject(&:empty?).join(", ")
-        suffix.empty? ? "" : ", #{suffix}"
-      end
-
-      def to_s
-        return "" unless compute_count
-        "#{@count} #{label}"
-      end
-
-      def as_json
-        return {} unless compute_count
-        { :"#{singular.gsub(' ', '_')}_count" => @count }
-      end
-
-    private
-
-      def compute_count
-        @count = block.call
-        !@count.nil?
-      rescue => e
-        RSpec.warning("End-of-run counter #{singular.inspect} raised: #{e.message}")
-        false
-      end
-
-      def label
-        @count.to_i == 1 ? singular : (plural || "#{singular}s")
-      end
-    end
-
     # rubocop:disable Metrics/ClassLength
 
     # Stores runtime configuration information.
@@ -462,8 +429,7 @@ module RSpec
       # @private
       attr_accessor :static_config_filter_manager
       # @private
-      attr_reader :backtrace_formatter, :ordering_manager, :loaded_spec_files,
-                  :end_of_run_counts
+      attr_reader :backtrace_formatter, :ordering_manager, :loaded_spec_files
 
       # rubocop:disable Metrics/AbcSize
       # rubocop:disable Metrics/MethodLength
@@ -523,7 +489,6 @@ module RSpec
         @world = World::Null
         @pending_failure_output = :full
         @force_line_number_for_spec_rerun = false
-        @end_of_run_counts = []
         @warnings = nil
 
         if warnings?
@@ -615,24 +580,6 @@ module RSpec
           add_setting(name, opts)
         end
         __send__("#{name}=", default) if default
-      end
-
-      # Registers a counter to be displayed in the end-of-run summary line.
-      # The block should return the current count when called.
-      #
-      # @param singular [String] singular label (e.g. "expectation")
-      # @param plural [String] optional plural label; defaults to singular + "s"
-      # @yield block that returns the current count
-      #
-      # @example
-      #     RSpec.configure do |c|
-      #       c.add_end_of_run_count("expectation") do
-      #         MyLib.expectation_count
-      #       end
-      #     end
-      def add_end_of_run_count(singular, plural=nil, &block)
-        raise ArgumentError, "A block that returns the count is required" unless block
-        @end_of_run_counts << EndOfRunCount.new(singular, plural, block)
       end
 
       # Returns the configured mock framework adapter module.
@@ -1000,6 +947,13 @@ module RSpec
             reporter.deprecation(*args)
           end
         end
+      end
+
+      # @private
+      # Returns the reporter if it has already been initialized,
+      # without triggering lazy initialization.
+      def initialized_reporter
+        @reporter
       end
 
       # @return [RSpec::Core::Reporter] the currently configured reporter
--- a/rspec-core/lib/rspec/core/formatters/html_printer.rb
+++ b/rspec-core/lib/rspec/core/formatters/html_printer.rb
@@ -57,12 +57,13 @@ module RSpec
                        "(PENDING: #{h(pending_message)})</span></dd>"
         end
 
-        def print_summary(duration, example_count, failure_count, pending_count)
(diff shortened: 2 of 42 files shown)


soulcutter:
I have a couple of thoughts:

* it doesn’t seem right to add a class-level ivar to expectations to support this count. RSpec is not the most parallelizable runner as it stands today, and this adds to that kind of architectural problem. This is likely not thread-safe as-is
* is EndOfRunCount intended to be used for the other counts as well?
* I wonder if instead of the block returning a number, if rspec could inject a shared counter object into integrating libraries (like expectations)… not sure how that would work out

I realistically won’t have time to review this deeply or give concrete suggestions for a while, I’m afk this week. It’s nice to see somebody pick this up.

rspec-core/lib/rspec/core/formatters/html_printer.rb
```diff
@@ -57,12 +57,13 @@ def print_example_pending(description, pending_message)
                        "(PENDING: #{h(pending_message)})</span></dd>"
         end
 
-        def print_summary(duration, example_count, failure_count, pending_count)
+        def print_summary(duration, example_count, failure_count, pending_count, end_of_run_counts=[])
```
JonRowe: This should take a single count, and add the required line

rspec-core/lib/rspec/core/formatters/json_formatter.rb
```diff
@@ -29,6 +29,7 @@ def dump_summary(summary)
             :pending_count => summary.pending_count,
             :errors_outside_of_examples_count => summary.errors_outside_of_examples_count
           }
+          summary.end_of_run_counts.each { |counter| @output_hash[:summary].merge!(counter.as_json) }
```
JonRowe: As above

rspec-core/lib/rspec/core/configuration.rb
```diff
@@ -6,6 +6,39 @@
 
 module RSpec
   module Core
+    # @private
+    # Value object representing a counter to display at end of run.
+    EndOfRunCount = Struct.new(:singular, :plural, :block) do
```
JonRowe: This doesn't belong here in my opinion, this is a function of the reporter, and I don't think this needs to be this complicated, a simple counter would be better.

rspec-core/lib/rspec/core/configuration.rb
```diff
@@ -582,6 +617,24 @@ def add_setting(name, opts={})
...
+      # @example
+      #     RSpec.configure do |c|
+      #       c.add_end_of_run_count("expectation") do
+      #         MyLib.expectation_count
+      #       end
+      #     end
+      def add_end_of_run_count(singular, plural=nil, &block)
+        raise ArgumentError, "A block that returns the count is required" unless block
+        @end_of_run_counts << EndOfRunCount.new(singular, plural, block)
+      end
```
JonRowe: This a function of formatters, and the core library should only record a simple count in the reporter, anything else should be up to their own library

rspec-core/lib/rspec/core/notifications.rb
```diff
@@ -283,10 +283,19 @@ def fully_formatted
...
                                      :pending_examples, :load_time,
-                                     :errors_outside_of_examples_count)
+                                     :errors_outside_of_examples_count,
+                                     :raw_end_of_run_counts)
     class SummaryNotification
+      # @api
+      # @return [Array<EndOfRunCount>] extra counters registered via Configuration#add_end_of_run_count
+      def end_of_run_counts
+        raw_end_of_run_counts || []
+      end
```
JonRowe: Making this simpler would negate the need to add extra "raw" fields.

rspec-expectations/lib/rspec/expectations.rb
```diff
@@ -83,5 +83,33 @@ class MultipleExpectationsNotMetError < ExpectationNotMetError
...
+
+      # @private
+      def increment_expectation_count
+        @expectation_count += 1
+      end
+
+      # @private
+      def reset_expectation_count
+        @expectation_count = 0
+      end
```
JonRowe: Rather than storing these here, they should call the reporter if available, this would also negate the need to add the configuration check below,

JonRowe (changes requested):
:wave: This is quite complex to add "custom counts" when all that is really needed is for the reporter to store a number of expectations, other counts would and should require additional formatter support and could run their own out of core expansion for this.

I think overall this should be a lot simpler to be included in core if at all...

rspec-core/lib/rspec/core/formatters/json_formatter.rb
```diff
@@ -29,6 +29,7 @@ def dump_summary(summary)
             :pending_count => summary.pending_count,
             :errors_outside_of_examples_count => summary.errors_outside_of_examples_count
           }
+          @output_hash[:summary][:expectation_count] = summary.expectation_count unless summary.expectation_count.nil?
```
JonRowe, line 32: If we're sending this we should always send it

rspec-core/lib/rspec/core/notifications.rb
```diff
@@ -318,6 +320,9 @@ def totals_line
             " occurred outside of examples"
           )
         end
+        if (expectation_count || 0) > 0
```
JonRowe, line 323: Don't be defensive here, ensure its 0 elsewhere.
```suggestion
        if expectation_count > 0
```

rspec-core/lib/rspec/core/formatters/html_printer.rb
```diff
@@ -57,10 +57,11 @@ def print_example_pending(description, pending_message)
                        "(PENDING: #{h(pending_message)})</span></dd>"
         end
 
-        def print_summary(duration, example_count, failure_count, pending_count)
+        def print_summary(duration, example_count, failure_count, pending_count, expectation_count=0)
```
JonRowe, line 60: Remove the defensive value here, it should always be sent.
```suggestion
        def print_summary(duration, example_count, failure_count, pending_count, expectation_count)
```

rspec-core/lib/rspec/core/formatters/html_printer.rb
```diff
@@ -57,10 +57,11 @@ def print_example_pending(description, pending_message)
...
           totals = "#{example_count} example#{'s' unless example_count == 1}, "
           totals << "#{failure_count} failure#{'s' unless failure_count == 1}"
           totals << ", #{pending_count} pending" if pending_count > 0
+          totals << ", #{expectation_count} expectation#{'s' unless expectation_count == 1}" if (expectation_count || 0) > 0
```
JonRowe, line 64: Remove the defensive checks here, it should always be sent and never be nil.
```suggestion
          totals << ", #{expectation_count} expectation#{'s' unless expectation_count == 1}"
```

rspec-core/lib/rspec/core/configuration.rb
```diff
@@ -949,6 +949,13 @@ def play_onto(reporter)
         end
       end
 
+      # @private
+      # Returns the reporter if it has already been initialized,
+      # without triggering lazy initialization.
+      def initialized_reporter
+        @reporter
+      end
```
JonRowe, line 957: Don't add this, the reporter will be initialised by the time you need it.

rspec-core/spec/rspec/core/notifications_spec.rb
```diff
@@ -452,6 +452,98 @@ module RSpec::Core::Notifications
...
+        it 'includes the expectation count in the summary line' do
+          expect(fully_formatted).to include('2 examples, 0 failures, 25 expectations')
+        end
+      end
+    end
+
+    describe '#totals_line' do
+      context "with no expectation count" do
+        it "returns the standard summary" do
+          expect(notification.totals_line).to eq("2 examples, 0 failures")
```
JonRowe, line 478: We always mention 0 failures, so we should also say 0 expectations

rspec-expectations/lib/rspec/expectations/handler.rb
```diff
@@ -21,6 +21,11 @@ def self.with_matcher(handler, matcher, message)
         ::RSpec::Matchers.last_matcher = matcher
       end
 
+      def self.notify_expectation
+        return unless ::RSpec.respond_to?(:configuration)
+        ::RSpec.configuration.initialized_reporter&.increment_expectation_count
```
JonRowe, line 26: This should already be initialised before use.
```suggestion
        return unless ::RSpec.respond_to?(:configuration)
        ::RSpec.configuration.reporter.increment_expectation_count
```

rspec-expectations/spec/rspec/expectations/handler_spec.rb
```diff
@@ -258,4 +258,19 @@ module Expectations
...
+    it "increments the reporter expectation count for a positive expectation" do
+      expect(reporter).to receive(:increment_expectation_count)
+      RSpec::Expectations::PositiveExpectationHandler.handle_matcher(true, matcher)
+    end
+
+    it "increments the reporter expectation count for a negative expectation" do
+      expect(reporter).to receive(:increment_expectation_count)
+      RSpec::Expectations::NegativeExpectationHandler.handle_matcher(false, matcher)
+    end
+  end
```
JonRowe, line 275: These specs should be reworked to use a real reporter and check the count increments, its not really adding value as is, this sort of means you might want to make the reporter configurable within the code or stub out just the reporter to a real one.

rspec-expectations/lib/rspec/expectations/handler.rb
tangposmarvin: Hi @JonRowe 

Thanks for the suggestion. I tried using `::RSpec.configuration.reporter.increment_expectation_count` directly. This works in production since the reporter is always initialized by the Runner before any expectations run.

However it causes issues in RSpec's own test suite. Calling `reporter` triggers [lazy
initialization](https://github.com/rspec/rspec/blob/6fd9256cfcb2db18be654f39fdecd7421b023122/rspec-core/lib/rspec/core/configuration.rb#L950-L968) which sets `@reporter`, and once set, [`output_stream=`](https://github.com/rspec/rspec/blob/6fd9256cfcb2db18be654f39fdecd7421b023122/rspec-core/lib/rspec/core/configuration.rb#L253-L262) refuses to accept a new value. This breaks in two places:

- **aruba [`run_command`](https://github.com/rspec/rspec/blob/6fd9256cfcb2db18be654f39fdecd7421b023122/rspec-core/spec/support/aruba_support.rb#L44-L71)** — after `RSpec.reset`, the outer
test's expectations initialize the reporter with `$stdout`. The next `run_command`'s
[`Runner#configure`](https://github.com/rspec/rspec/blob/6fd9256cfcb2db18be654f39fdecd7421b023122/rspec-core/lib/rspec/core/runner.rb#L129-L133) tries to set `output_stream = temp_stdout`
but it's rejected, so inner run output goes to `$stdout` and `last_cmd_stdout` is empty.

- **bisect specs** — `expect { bisect_call }.to avoid_outputting` initializes the reporter before `Runner#configure` can set the correct streams, causing a warning to stderr which fails
the `avoid_outputting` check.

A couple of approaches I considered:

1. **Call `reporter` directly** — requires clearing `@reporter` in `Runner#configure` before setting streams, so the reporter gets re-created with the correct streams. Small production
code change.

2. **Store the count in rspec-expectations, read it from Reporter at finish** — rspec-expectations increments a simple counter, and [`Reporter#finish`](https://github.com/rspec/rspec/blob/6fd9256cfcb2db18be654f39fdecd7421b023122/rspec-core/lib/rspec/core/reporter.rb#L179-L198) reads it when building the `SummaryNotification`. Same pattern as [`@non_example_exception_count`](https://github.com/rspec/rspec/blob/6fd9256cfcb2db18be654f39fdecd7421b023122/rspec-core/lib/rspec/core/reporter.rb#L171). No cross-library `reporter` call, no test infrastructure changes.

How would you like me to handle this?

Issue #13, opened 2015-08-06
Print test failures on INFO signal


tenderlove:
Hi,

One thing I miss from minitest is if I have a slow test suite, and there is a failure, I can hit Ctrl-T to send a SIGINFO and it will print out the current failures with a stack trace.  Then I can decide if I want to kill the process and fix those or just let it go.

I made [a video to demonstrate](http://youtu.be/ovOPbzLT0e8).

I will try to make a PR for this, but I wanted to file a ticket a) so I don't forget, and b) in case someone can get to it before me.

Thanks!

PR #314, opened 2026-03-04
Print threads backtrace on SIGINFO/SIGPWR

sobrinho:
This can help to get stuck build backtraces.

For example in a CI:

```ruby
pid = spawn("bundle exec rspec")
started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)

monitor = Thread.new do
  while Process.kill(0, pid)
    if (Process.clock_gettime(Process::CLOCK_MONOTONIC) - started_at) > 60 # seconds
      Process.kill("INFO", pid)
    end

    sleep 5
  end
end

_, status = Process.waitpid2(pid)
monitor.kill

exit(status.exitcode)
```

For local development, you can also do it manually by getting the PID from `ps aux` or by doing this:

```bash
bundle exec rspec &
pid=$!
kill -INFO $!
```

_This is a common pattern used by [Sidekiq](https://github.com/sidekiq/sidekiq/blob/c2941506ddf5c8b691d0c6c318f454a76b0abf78/lib/sidekiq/cli.rb#L216-L225) and [Puma](https://github.com/puma/puma/blob/a1b5b5e7e1b8d34b7d24964f668733299be930a2/lib/puma/launcher.rb#L478-L489) that I'm aware of but possibly other projects as well._


sobrinho:
This also converges to adding this kind of stuff to other projects such as https://github.com/grosser/parallel_tests/issues/1028.

By having a script like the one proposed here, adding the SIGINFO there as well, we have an end-to-end solution to debug stuck builds automatically on CI when it is flaky.

sobrinho:
I tried to make a constant instead of a method for `backtrace_signal` but then the SimpleCov complained because only the branch supported by the Ruby implementation is tested.

Since this is only called to setup and it is not a hotspot, using a method should be fine.

Screenshot 2026-03-03 at 22 10 37

rspec-core/spec/rspec/core/runner_spec.rb
```diff
@@ -155,6 +155,99 @@ def interrupt
       end
     end
 
+    describe "backtrace handling" do
+      it "prints the backtrace of main thread" do
+        # JRuby names the main thread "main", so we need to stub it to get a consistent output.
```
JonRowe, line 160: If this is the case it should only be stubbed on JRuby

JonRowe (commented):
👋 thanks for the submission, there is prior discussions about signal handling in #8 #13 and rspec/rspec-core#741

Something that would need to be added here is handling of existing interrupts when `trap` is called, if you wanted/were able to add that to the existing handler at the same time that would be great.

Another is a cucumber scenario testing this or an integration test without mocking...

rspec-core/lib/rspec/core/runner.rb
```diff
@@ -176,6 +177,35 @@ def self.trap_interrupt
...
+        trap(backtrace_signal) { handle_backtrace } if backtrace_signal
+      end
+
+      # @private
+      def self.backtrace_signal
+        if Signal.list.key?("INFO")
+          "SIGINFO"
+        elsif Signal.list.key?("PWR")
+          "SIGPWR"
+        end
```
JonRowe, line 191: If you want to make this a constant you can use `:nocov:` to make single cov ignore the uncalled case.

rspec-core/spec/rspec/core/runner_spec.rb
sobrinho: This makes the spec resilient to future changes on this. If MRI ever decides to name the thread or other Ruby implementations uses a name, that will make the spec consistent everywhere.

The thread being named is spec'ed in the other scenarios.

rspec-core/lib/rspec/core/runner.rb
sobrinho: That should be fine, we don't expect this method to be called more than once on boot anyways.

sobrinho:
@JonRowe I don't think we should call the original SIGPWR/SIGINFO. Why would we? Aren't we trapping the signal to dump the thread backtraces?

sobrinho:
@JonRowe how do you feel about this cucumber test?

sobrinho:
This is not working as expected on Linux, I'm working on it.

sobrinho:
It looks good to me now.

Would you review again, @JonRowe?

rspec-core/features/step_definitions/additional_cli_steps.rb
```diff
@@ -219,6 +219,11 @@
   step "I run `#{cmd}`"
 end
 
+When(/^I run `([^`]+)` with the backtrace signal sent during the run$/) do |cmd|
+  set_environment_variable('RUBYOPT', (ENV['RUBYOPT'] || '') + " -r#{File.expand_path("../../support/send_backtrace_signal.rb", __FILE__)}")
+  step "I run `#{cmd}`"
```
luke-hill, line 224: Not one for now, and I can see it above. But one thing to be mindful of is this is an antipattern, one we committed to removing a long time ago but never got around to

Maybe one for @JonRowe just to keep in the backlog. As/when we get round to this I promise to give an update

JonRowe: Can you clarify what is considered an antipattern here? `step`?

luke-hill: Sorry yes. This is something that was slated for removal about 5 years ago but we never got around to it.

https://github.com/cucumber/cucumber-ruby/issues/1362

sobrinho: I don't think we should worry about this right now. We have other usages of `step` in the codebase and others mentioned in the issue this feature is quite useful.

We can deal with it if it ever gets removed from Cucumber.

sobrinho:
@JonRowe would you review again?

sobrinho:
@JonRowe ping

sobrinho:
@JonRowe ping 👀

sobrinho:
@JonRowe anything blocking this PR?

PR #315, opened 2026-03-05
Document early-stopping behaviour of mocks

amomchilov:
Mocked methods like `.never` usually end the execution of the code under test, but not always.

```ruby
def get_name = "Bob"
def greet(name) = puts "Hello #{name}"

expect(self).to receive(:get_name).never

begin
  name = get_name # This raises `MockExpectationError` immediately
  greet(name) # Not called
rescue Exception => e
  # swallowed (but RSpec still knows it happened, and will fail the example when it finishes)
end

puts "but this still gets called"
```

This PR documents this behaviour, which might be surprising.

### DIFF
--- a/rspec-mocks/lib/rspec/mocks/error_generator.rb
+++ b/rspec-mocks/lib/rspec/mocks/error_generator.rb
@@ -4,6 +4,16 @@ RSpec::Support.require_rspec_support 'ruby_features'
 module RSpec
   module Mocks
     # Raised when a message expectation is not satisfied.
+    #
+    # This inherits from `Exception` (not `StandardError`) so that a
+    # `rescue => e` or `rescue StandardError` in the code under test
+    # will not accidentally swallow it. This allows rspec to immediately
+    # halt execution of the tested code when an unexpected message is received.
+    #
+    # However, code that does `rescue Exception` *will* intercept this error.
+    # The failure will still be reported at the end of the example,
+    # but the tested code will continue executing instead of stopping immediately,
+    # which may cause confusing behavior or misleading secondary errors.
     MockExpectationError = Class.new(Exception)
 
     # Raised when a test double is used after it has been torn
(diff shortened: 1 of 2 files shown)


pirj:
How does it work with https://rspec.info/features/3-13/rspec-core/expectation-framework-integration/aggregating-failures/?

rspec-mocks/lib/rspec/mocks/error_generator.rb
```diff
@@ -4,6 +4,16 @@
...
+    #
+    # This inherits from `Exception` (not `StandardError`) so that a
+    # `rescue => e` or `rescue StandardError` in the code under test
+    # will not accidentally swallow it. This allows rspec to immediately
+    # halt execution of the tested code when an unexpected message is received.
+    #
+    # However, code that does `rescue Exception` *will* intercept this error.
+    # The failure will still be reported at the end of the example,
+    # but the tested code will continue executing instead of stopping immediately,
+    # which may cause confusing behavior or misleading secondary errors.
```
JonRowe, line 16:
```suggestion
    #
    # Note: this inherits from `Exception` (rather than `StandardError`) in order to
    # minimise the chance that might accidentally be rescued. However, code that does
    # `rescue Exception` *will* intercept this error and prevent RSpec from short
    # circuiting the failure. Whilst the failure will still be reported at the end of
    # the example, the code under test will continue executing which may cause
    # misleading secondary errors or confusing behaviour.
```

rspec-mocks/lib/rspec/mocks/message_expectation.rb
```diff
@@ -230,6 +230,10 @@ def and_yield(*args, &block)
       # Constrain a message expectation to be received a specific number of
       # times.
       #
+      # If the message is received too many times, a {MockExpectationError}
+      # will be raised immediately. This usually stops the rest of the example
+      # from running, but not always. See the {MockExpectationError} documentation.
+      #
```
JonRowe, line 236: I would shorten all of these just adding the reference to `See the {MockExpectationError} documentation.` After the first message.

JonRowe (changes requested):
Happy to tweak the docs here I've made some suggestions.

PR #322, opened 2026-04-28
chore: correct failing cops

chadlwilson:
This reverts commit 8b96620688e9e2dd1fe5834ec26646e839285854.

I guess the upstream change was since addressed.


JonRowe:
Thanks... 🤷 it passes so... I guess too

Issue #321, opened 2026-04-28
JRuby 10.1 can be subject to infinite recursion during `:respond_to?` stubbing


chadlwilson:
I've been playing around ensuring the tests pass on JRuby 10.1 (CRuby 4.0 compatible). Generally looking pretty good[1] however, the workaround for infinite recursion at https://github.com/rspec/rspec/commit/939c4799993b7ff7e524fac701ae6490772ca6de doesn't work on JRuby and gives a `StackOverflowError` when trying to compute the error message.

This one sometimes fails
https://github.com/rspec/rspec/blob/76f3447e978a18b216ddf03f1fa360d3b695959d/rspec-mocks/spec/rspec/mocks/matchers/receive_spec.rb#L614

This is likely because the `caller_location` `label` is `inspect` rather than `Kernel#inspect` on JRuby[2]. It seems JRuby never implemented the same prefixing as CRuby introduced in Ruby 3.4.

https://github.com/rspec/rspec/blob/8526897e6bc32b35e2ba307a3d6327ffdf44b33d/rspec-mocks/lib/rspec/mocks/method_double.rb#L73-L74

I'm not sure of an appropriate workaround here; since checking `caller_location[0].label == "inspect"` seems a bit "loose" and might have other consequences, so raising this for discussion.

Forgive my bad test code
```ruby
class InspectShowsCaller
  def inspect
    "inspect caller label is [#{caller_locations[0].label}]"
  end
end

class DefaultInspect
  def initialize
    @dodgy = InspectShowsCaller.new
  end
end

puts "Testing #{RUBY_VERSION} / #{defined?(JRUBY_VERSION) ? JRUBY_VERSION : 'CRuby'}: #{DefaultInspect.new.inspect}"
```

```
$ for v in 3.1 3.2 3.3 3.4 4.0 jruby-9.4 jruby-10.0 jruby-10.1; do mise x ruby@$v -- ruby check-caller-location.rb; done
Testing 3.1.7 / CRuby
#<DefaultInspect:0x000000010cff6bc8 @dodgy=inspect caller label is [inspect]>
Testing 3.2.9 / CRuby
#<DefaultInspect:0x0000000109c31e00 @dodgy=inspect caller label is [inspect]>
Testing 3.3.11 / CRuby
#<DefaultInspect:0x00000001033a8208 @dodgy=inspect caller label is [inspect]>
Testing 3.4.8 / CRuby
[...]

PR #323, opened 2026-04-28
test: test against JRuby 10.1

chadlwilson:
JRuby 10.1 is CRuby 4.0 compatible and has been released: https://www.jruby.org/2026/04/21/jruby-10-1-0-0.html

This PR
- enables tests for 10.1
- pends/skips a few failing specs; linking to their relevant issues
- removes the old spec/test hacks for JRuby 9.1 - this has been long since EOL, and running specs on JRuby I believe is best efforts anyway so don't think maintainer should need to support these.

Needs investigation subsequently:
- https://github.com/rspec/rspec/issues/321 (might be at least partly a JRuby issue)
- https://github.com/jruby/jruby/issues/9398

### DIFF
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -56,8 +56,9 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - jruby-9.4
-          - jruby-10.0
+          - jruby-10.1 # ruby 4.0 compatible
+          - jruby-10.0 # ruby 3.4 compatible
+          - jruby-9.4  # ruby 3.1 compatible
       fail-fast: false
     continue-on-error: false
     defaults:
--- a/rspec-core/spec/integration/bisect_spec.rb
+++ b/rspec-core/spec/integration/bisect_spec.rb
@@ -15,12 +15,6 @@ module RSpec::Core
       normalize_durations(formatter_output.string)
     end
 
-    before do
-      if RSpec::Support::Ruby.jruby? && RSpec::Support::Ruby.jruby_version == '9.1.17.0'
-        skip "These specs are currently broken on JRuby 9.1.17.0"
-      end
-    end
-
     context "when a load-time problem occurs while running the suite" do
       it 'surfaces the stdout and stderr output to the user' do
         output = bisect(%w[spec/rspec/core/resources/fail_on_load_spec.rb_], 1)
--- a/rspec-core/spec/integration/spec_file_load_errors_spec.rb
+++ b/rspec-core/spec/integration/spec_file_load_errors_spec.rb
@@ -236,7 +236,10 @@ RSpec.describe 'Spec file load errors' do
         end
       end
     else
-      it 'prints a basic error when no syntax_suggest is available/loaded', :skip => RSpec::Support::Ruby.jruby? do
+      it 'prints a basic error when no syntax_suggest is available/loaded' do
+        if RSpec::Support::Ruby.jruby? && RSpec::Support::Ruby.jruby_version < '10'
+          skip 'JRuby < 10 cannot get the __send__ backtrace correctly'
+        end
         run_command "--require ./broken_file"
         expect(last_cmd_exit_status).to eq(error_exit_code)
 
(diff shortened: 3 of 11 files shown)


JonRowe:
I am happy to merge this as is but I'd like to ask you as someone who uses JRuby, whats the trade off here for instead dropping 10.0? Do people regularly upgrade to the latest point release? I want to keep our JRuby suite fairly minimal becauses its historically been a big drain maintenance wise.

JonRowe (approved):

chadlwilson:
Understood. I think it's probably uncharted ground since JRuby skipped ruby 3.2 and 3.3 compatibility, so there is/was a big jump getting to JRuby 10.0 and its ruby 3.4 compatibility, which also requires Java 21. Having 4.0 compatibility within a few months of CRuby release is thus a pretty big achievement. I imagine for those on 10.0 moving to 10.1 won't be so difficult, especially as ruby 3.4 to 4.0 wasn't so big.

JRuby 9.4 is "soft" EOL now, basically commercial support only. So you could consider dropping actively testing against 9.4 if you're trying to rationalize. I don't really have a particular view on whether it'd be better to drop 9.4 or 10.0 for the ecosystem, as I don't have any real data to go on.

chadlwilson:
If it helps with the maintenance concerns, I had a subsequent PR where I was going to suggest dropping various workarounds/hacks for older JRubies from the production code.

JonRowe:
> I don't really have a particular view on whether it'd be better to drop 9.4 or 10.0 for the ecosystem, as I don't have any real data to go on.

For clarification, I meant keeping 9.4 and 10.1 as the two major branches.

chadlwilson:
Yup, I understood. My personal view is that id drop 9.4 before 10.0, because it's basically EOL.

I just don't know if dropping 10.0 makes sense right now, since 10.1 has only been out for a couple of weeks so there is zero experience with difficulty of migration, and even 10.0 has only been out for 1 year with a 3-year CRuby version jump from 9.4 to that, so a huge lag behind.

To my knowledge there is not a comparable point release that came so quickly between JRuby versions earlier. From my perspective there were reasonably big differences between 9.2, 9.3 and 9.4 which is as far back as my history goes, often with the gem compatibility nightmare from other libraries rather than JRuby itself. I still havent migrated my main project I support even to 10.0 because the Java 21 minimum is a huge jump,.and there are all sorts of implications for rails.

is the cost of keeping JRuby tests working that high? Im a bit confused since it's still testing the expensive tests against CRuby 3.0, 3.1 and 3.2 which are all even security EOL (albeit much larger audiences).

JonRowe:
> is the cost of keeping JRuby tests working that high? 

JRuby is probably more stable these days but it has historically been high maintenance for RSpec, and much slower to test at parity with CRuby due to start up costs (due to how our test suite is written) its why they run a cut down set at the moment. I want to keep a sanity check but I don't have the time to invest in every little JRuby difference myself these days.

chadlwilson:
Yeah, fair enough. It's pretty stable these days (outside Windows problems) possibly because CRuby itself has focused more on areas that don't affect JRuby so much for a period. There is also a JRuby prism implementation via WASM now too - noticed you've been doing some prism stuff.

I have seen that the the current reentrant mutex test flakes on JRuby sometimes, so there is something there that is probably not guaranteed on JRuby (maybe something in ThreadOrder?)

If there are issues going forward you can skip/pend the tests and tag me in an issue and I'll see whether it's within my capability to resolve without blocking you.

I might set up something scheduled in my fork as well.

chadlwilson:
The startup speed has improved a lot due to distribution of shared archives (essentially precompiled bytecode caches) and improved Java defaults to enable use of these.

I didn't actively compare the cucumber speeds between CRuby and JRuby but on my overpowered dev machine it didn't seem too bad 😅

JonRowe:
> The startup speed has improved a lot due to distribution of shared archives (essentially precompiled bytecode caches) and improved Java defaults to enable use of these.

Its not JRubies fault, its just the way we start so many cli runs, and for example we test spec files in isolation on CRuby which is a start up per spec file... 

> I didn't actively compare the cucumber speeds between CRuby and JRuby but on my overpowered dev machine it didn't seem too bad 😅

I'd point to an old pre-monorepo build to highlight the difference, but it seems those logs have been "tidied" by Github since archival.

chadlwilson:
Yeah, I'd be interested to see what it's like these days - might have a go one day. The startup speed of JRuby and the JVM is likely a big difference, and still a downside of JRuby, but it's certainly gotten better on 10.0 and onwards even without JVM startup args incantations (like `--dev` to jruby.sh does).

PR #326, opened 2026-05-04
test: correct tests for Aruba 2.4.0 / Cucumber 11 support

### LINKED ISSUE #325: Tests failing on `main` due to Aruba 2.4.0 release
Aruba 2.4.0 now brings in a dependency on irb due to https://github.com/cucumber/aruba/pull/969, which brings a whole lot of other stuff, including bundled `rdoc`, `erb`, `psych`, `tsort`, `pp` etc:

https://github.com/cucumber/aruba/compare/v2.3.3...v2.4.0#diff-28c3d98952a8fbb31760941c648b1fe684a42c394422f6c8b9937cb97f202a54

https://github.com/chadlwilson/rspec/actions/runs/25275735223/job/74116339002

Among other things
- this changes backtraces which were perhaps assuming default gem `erb`and thus seems to break formatter tests which now expect various `erb` lines differing between Ruby versions.
- somehow caused an unrelated anonymous module to appear breaking some jruby specs
- broke some Ruby 3.0 cucumber tests due to confusing tsort versions to activate (now tsort `0.1.0` vs `0.2.0`)
- library wide checks needs to allow `set.rb` on Ruby < 4.0 (?)

Not sure if the correct thing to do is just change all the test expectations (since then they will depend on aruba 2.4.0 and its transitive dependencies), so just raising this for now.

chadlwilson:
fixes #325

As discussed in #325, Aruba 2.4.0 adds a dependency on `irb` which in turn brings in Gemified versions of libraries that otherwise were being loaded from default gems or differently across Ruby versions.

The failing tests prior to these changes can be seen at https://github.com/chadlwilson/rspec/actions/runs/25275735223/job/74116339002

I chose the versions such that they should be OK for easier Rspec `3.13` backport but I note that the `Gemfile` has had some refactoring on `main` for `4.x` so that might not be a valid consideration. I'll add comments below.

I'm also happy to raise a PR for 3.13 backport to get tests passing again there too.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,10 +19,6 @@ group :documentation do
   gem 'yard', '~> 0.9.24', :require => false
 end
 
-platforms :jruby do
-  gem "jruby-openssl"
-end
-
 #
 # Support for gems extracted in Ruby versions
 #
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,7 @@ end
 gem 'bigdecimal', :require => false if RUBY_VERSION.to_f >= 3.3
 gem 'drb' if RUBY_VERSION.to_f >= 3.3
 gem 'mutex_m', '~> 0.1.0' if RUBY_VERSION.to_f > 3.3
+gem 'tsort', '0.1.0' if RUBY_VERSION.to_f < 3.1
 
 #
 # Tooling / our testing
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,6 @@ end
 
 gem 'bigdecimal', :require => false if RUBY_VERSION.to_f >= 3.3
 gem 'drb' if RUBY_VERSION.to_f >= 3.3
-gem 'mutex_m', '~> 0.1.0' if RUBY_VERSION.to_f > 3.3
 gem 'tsort', '0.1.0' if RUBY_VERSION.to_f < 3.1
 
 #
@@ -49,7 +48,7 @@ gem "rubocop", "~> 1.80" if RUBY_VERSION >= '3.3' && RUBY_ENGINE == 'ruby'
 #
 
 gem 'flexmock', '~> 0.9.0'
-gem 'minitest', '~> 5.12.0'
+gem 'minitest', '~> 5.15'
 gem 'mocha', '~> 0.13.0'
 gem 'rr', "~> 1.0.4"
 gem 'test-unit', '~> 3.0'
--- a/rspec-core/spec/rspec/core/shared_example_group_spec.rb
+++ b/rspec-core/spec/rspec/core/shared_example_group_spec.rb
@@ -161,8 +161,9 @@ module RSpec
               )).and exclude(a_string_matching(anonymous_module_regex))
             )
 
-            expect(example_group.ancestors.map(&:inspect)).to include_a_named_rather_than_anonymous_module
-            expect(example_group.ancestors.map(&:to_s)).to include_a_named_rather_than_anonymous_module
+            relevant_ancestors = example_group.ancestors.take_while { |x| x != RSpec::Core::MemoizedHelpers }
+            expect(relevant_ancestors.map(&:inspect)).to include_a_named_rather_than_anonymous_module
+            expect(relevant_ancestors.map(&:to_s)).to include_a_named_rather_than_anonymous_module
           end
 
           ["name", :name, ExampleModule, ExampleClass].each do |object|
--- a/rspec-core/spec/support/formatter_support.rb
+++ b/rspec-core/spec/support/formatter_support.rb
@@ -34,6 +34,7 @@ module FormatterSupport
     configuration = runner.configuration
     configuration.filter_gems_from_backtrace "gems/bundler"
     configuration.backtrace_formatter.exclusion_patterns << /rspec_with_simplecov/
+    configuration.backtrace_formatter.exclusion_patterns << /erb\.rb/ # remove differences/churn between use of stdlib and bundled erb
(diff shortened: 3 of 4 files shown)


Gemfile
```diff
@@ -48,7 +44,7 @@ gem "rubocop", "~> 1.80" if RUBY_VERSION >= '3.3' && RUBY_ENGINE == 'ruby'
 #
 
 gem 'flexmock', '~> 0.9.0'
-gem 'minitest', '~> 5.12.0'
+gem 'minitest', '~> 5.15'
```
chadlwilson, line 47: `5.15` is the last version that declares Ruby 2.2 support as would be needed for Rspec 3.13 tests, but not sure if that actually works correctly. Later versions fix `mutex_m` require issues on those newer Ruby versions.

This could equally be `~> 5.26` to support Ruby 3.0 -> 4.0 - maybe better to declare that than guess for `3.13` branch here. Hmm.

rspec-core/spec/rspec/core/shared_example_group_spec.rb
```diff
@@ -161,8 +161,9 @@ def find_implementation_block(registry, scope, name)
               )).and exclude(a_string_matching(anonymous_module_regex))
             )
 
-            expect(example_group.ancestors.map(&:inspect)).to include_a_named_rather_than_anonymous_module
-            expect(example_group.ancestors.map(&:to_s)).to include_a_named_rather_than_anonymous_module
+            relevant_ancestors = example_group.ancestors.take_while { |x| x != RSpec::Core::MemoizedHelpers }
+            expect(relevant_ancestors.map(&:inspect)).to include_a_named_rather_than_anonymous_module
+            expect(relevant_ancestors.map(&:to_s)).to include_a_named_rather_than_anonymous_module
```
chadlwilson, line 166: As in the commit message:

> For one reason or another PP itself sometimes ends up with anonymous modules in the ancestors on JRuby 10.0+ once
pp is using a bundled version (pulled in via aruba 2.4.0 -> irb -> pp). These older parts of the ancestors
don't seem relevant to the test, so stop looking once we get to our own MemoizedHelpers. Probably some magic JRuby
is doing somehow?

eg
```
#...
"RSpec::Core::Pending",
"RSpec::Core::MemoizedHelpers",
"#<Module:0x6eafcfc8>",
"Object",
"JSON::Ext::Generator::GeneratorMethods::Object",
"PP::ObjectMixin",
"#<Module:0x54ad833>",
"Kernel",
"BasicObject"]
```

rspec-core/spec/support/formatter_support.rb
```diff
@@ -34,6 +34,7 @@ def run_rspec_with_formatter(formatter, options={})
     configuration = runner.configuration
     configuration.filter_gems_from_backtrace "gems/bundler"
     configuration.backtrace_formatter.exclusion_patterns << /rspec_with_simplecov/
+    configuration.backtrace_formatter.exclusion_patterns << /erb\.rb/ # remove differences/churn between use of stdlib and bundled erb
```
chadlwilson, line 37: I think these are OK to exclude for purpose of the formatter specs since the formatter uses `erb`?

rspec-support/spec/rspec/support_spec.rb
```diff
@@ -11,7 +11,7 @@ module RSpec
       :consider_a_test_env_file => %r{rspec/support/spec},
       :allowed_loaded_feature_regexps => [
         /rbconfig/, # Used by RubyFeatures
-        /prettyprint.rb/, /pp.rb/, /diff\/lcs/ # These are all loaded by the differ.
+        /prettyprint\.rb/, /pp\.rb/, /set\.rb/, /diff\/lcs/ # These are all loaded by the differ.
```
chadlwilson, line 14: Couldn't find exactly which part is using `Set`, but these seem to be loaded/required somewhere within newer bundled `pp` versions or differ code.

> Set is a quasi-core gem Ruby 3.2 -> 3.4 and will seemingly show up as loaded on newer pp versions when pp itself is bundled.

Gemfile
```diff
@@ -19,17 +19,13 @@ group :documentation do
   gem 'yard', '~> 0.9.24', :require => false
 end
 
-platforms :jruby do
-  gem "jruby-openssl"
-end
-
```
chadlwilson, line 25: Just simplifying things here - using the bundled version is probably OK with JRuby, as it will have been exercised through even more tests (from CRuby test suite for openssl) than the latest standalone gem version.

Gemfile
```diff
@@ -19,17 +19,13 @@ group :documentation do
...
 #
 # Support for gems extracted in Ruby versions
 #
 
 gem 'bigdecimal', :require => false if RUBY_VERSION.to_f >= 3.3
 gem 'drb' if RUBY_VERSION.to_f >= 3.3
-gem 'mutex_m', '~> 0.1.0' if RUBY_VERSION.to_f > 3.3
+gem 'tsort', '0.1.0' if RUBY_VERSION.to_f < 3.1
```
chadlwilson, line 28: Matching the tsort stdlib version included with Ruby 3.0 since it seemingly does not support the later gemification changes of tsort until 3.1 and multiple versions are attempted to be activated.

mutex_m isn't needed to be manually added once using later gem versions elsewhere as they declare their dependencies properly for modern Rubies.

JonRowe:
Thanks for this!

chadlwilson:
Is the 3.13 branch build still OK? If not, would you like me to create a backport PR for that?

PR #328, opened 2026-05-19
feat(mocks): verify constant names in stub_const

### LINKED ISSUE #232: Make `stub_const` optionally check for constant presence
Would you be open to a PR to add verifying `stub_const`? eg:

```ruby
class User
  OFFER_VALUE = 3
end

# spec tests ->

it "reads the value" do
  stub_const_verifying("User::OFFER_VALUATION", 4)
  # -> ArgumentError:
  Cannot stub constant User::OFFER_VALUATION because it is not defined
end
```

#### Rational

We strongly favour a fail-fast startup in our apps.

eg, we use this to ensure our ENV is all correct before proceeding with app boot proper:

```ruby
class User
  OFFER_VALUE = `ENV.fetch("OFFER_VALUE")`
end
```

If we then want to test various different states of ENV, we need to stub const, but if the const name is mistyped, we get no feedback.

Adding this as a new method would be trivial. Would you like a PR for it?

Maybe there could be an rspec config which says "the standard `stub_const` is now verifying, wherever used"

Any ideas or suggestions gratefully received!

SAY-5:
Closes #232. When `verify_doubled_constant_names` is enabled, `stub_const` now raises if given the name of an undefined constant, matching the behaviour of `instance_double`/`class_double`. Following @JonRowe's note that he is "open to a PR that makes stub_const check verify_doubled_constant_names and raise if not set" and that this should be opt-in, the check only fires when the config option is set, so existing behaviour is unchanged by default.

### DIFF
--- a/rspec-mocks/lib/rspec/mocks/configuration.rb
+++ b/rspec-mocks/lib/rspec/mocks/configuration.rb
@@ -50,10 +50,11 @@ module RSpec
       end
 
       # When this is set to true, an error will be raised when
-      # `instance_double` or `class_double` is given the name of an undefined
-      # constant. You probably only want to set this when running your entire
-      # test suite, with all production code loaded. Setting this for an
-      # isolated unit test will prevent you from being able to isolate it!
+      # `instance_double`, `class_double` or `stub_const` is given the name of
+      # an undefined constant. You probably only want to set this when running
+      # your entire test suite, with all production code loaded. Setting this
+      # for an isolated unit test will prevent you from being able to isolate
+      # it!
       attr_writer :verify_doubled_constant_names
 
       # Provides a way to perform customisations when verifying doubles.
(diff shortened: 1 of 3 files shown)


rspec-mocks/lib/rspec/mocks/configuration.rb
```diff
@@ -56,6 +57,31 @@ def verify_doubled_constant_names?
...
+      # * `:namespace` - an error is raised when the constant's enclosing
+      #   namespace is not defined (e.g. stubbing `Foo::Bar` when `Foo` does
+      #   not exist).
+      # * `:full` - an error is raised when the constant itself is not already
+      #   defined.
+      #
+      # You probably only want `:full` when running your entire test suite,
+      # with all production code loaded; it will prevent you from stubbing
+      # constants in isolated unit tests.
+      def verify_constant_names=(value)
```
JonRowe: I would be more explicit about the name here and call it `verify_stub_constant_mode` as it will only apply to the `stub_const` edge cases and shouldn't be confused with the other verifying control

JonRowe (changes requested):

rspec-mocks/lib/rspec/mocks/configuration.rb
SAY-5: Renamed it to `verify_stub_constant_mode` throughout (config, error messages, and specs); the 173 mutate_const examples still pass.

PR #334, opened 2026-08-06
Fix NoMethodError when including a shared group with metadata into an example

### LINKED ISSUE #333: Including a metadata-carrying shared context into an example raises `NoMethodError: undefined method 'key?' for nil`
## Summary

When `config.include_context` is used to auto-include a **shared context that carries its own metadata** into examples matching some filter, and a matched **example** (not its group) also carries a metadata key that the shared context declares, RSpec crashes while configuring that example:

```
NoMethodError:
  undefined method 'key?' for nil
  rspec-core-3.13.6/lib/rspec/core/example_group.rb:736:in 'block in RSpec::Core::ExampleGroup.update_inherited_metadata'
```

The crash is surfaced to users as a confusing, unrelated timing error, because it happens after the example body runs but before the timing is recorded:

```
TypeError:
  can't convert NilClass into an exact number
  rspec-core-3.13.6/lib/rspec/core/example.rb:618:in 'RSpec::Core::Example::ExecutionResult#calculate_run_time'
```

Every affected example reports this `TypeError`, so the real cause (`NoMethodError` on `nil`) only shows up in the `--- Caused by: ---` section of the backtrace.

## Environment

- rspec-core 3.13.6 (rspec 3.13.2, rspec-support 3.13.7)
- Ruby 4.0.0
- Also reproduces on earlier 3.13.x.

## Reproduction

`config.shared_context_metadata_behavior = :apply_to_host_groups` is **required** to trigger this — with the legacy default (`:trigger_inclusion`) the bug does not occur.

```ruby
require "bundler/inline"

gemfile do
  gem "rspec-core", "3.13.6"
end

require "rspec/autorun"

RSpec.configure do |config|
  # Required: with the legacy default (:trigger_inclusion) the bug does not occur.
  config.shared_context_metadata_behavior = :apply_to_host_groups

  # A shared context that carries its OWN metadata (`:needs_setup`).
  RSpec.shared_context "shared setup", :needs_setup do
    let(:helper) { :helper }
  end

  # Auto-include that context into anything tagged `:use_setup`.
  config.include_context "shared setup", :use_setup
end

RSpec.describe "a group" do
  # This EXAMPLE (not the group) is tagged `:use_setup`, so RSpec includes
  # "shared setup" into the example's *singleton* example group at run time.
  #
  # It is also tagged `:needs_setup` -- the same key the shared context carries --
  # so that when RSpec propagates the context's metadata via
  # `update_inherited_metadata`, the key collides and the merge block runs.
  # That block dereferences `@user_metadata`, which is `nil` for a singleton
  # example group, raising `NoMethodError: undefined method 'key?' for nil`.
  it "an example", :use_setup, :needs_setup do
  end
end
```

```
$ ruby rspec_singleton_metadata_bug.rb
F

Failures:

  1) a group an example
     Failure/Error: self.run_time = (finished_at - started_at).to_f

     TypeError:
       can't convert NilClass into an exact number
     ...
     # --- Caused by: ---
     # NoMethodError:
     #   undefined method 'key?' for nil
     #   .../rspec/core/example_group.rb:736:in 'block in RSpec::Core::ExampleGroup.update_inherited_metadata'

1 example, 1 failure
```

(Standalone file: `rspec_singleton_metadata_bug.rb`, attached.)

## Expected behavior

The example runs and passes. Auto-including a metadata-carrying shared context into an example via a filter should not crash.

## AI-Investigated Root cause

At example run time, `Configuration#configure_example` includes any filter-matched modules into the example's **singleton example group** (the anonymous group that exists for a single example):

```ruby
# rspec/core/configuration.rb
def configure_example(example, example_hooks)
  example_hooks.register_global_singleton_context_hooks(example, hooks)
  singleton_group = example.example_group_instance.singleton_class

  singleton_group.with_replaced_metadata(example.metadata) do
    modules = @include_modules.items_for(example.metadata)
    modules.each do |mod|
      safe_include(mod, example.example_group_instance.singleton_class)
    end
    ...
  end
end
```

When `mod` is a `SharedExampleGroupModule` that carries metadata, `included` propagates that metadata to the host via `update_inherited_metadata`:

```ruby
# rspec/core/shared_example_group.rb
def include_in(klass, inclusion_line, args, customization_block)
  klass.update_inherited_metadata(@metadata) unless @metadata.empty?
  ...
end
```

`update_inherited_metadata` reads `@user_metadata` in its merge block:

```ruby
# rspec/core/example_group.rb:734
def self.update_inherited_metadata(updates)
  metadata.update(updates) do |key, existing_group_value, new_inherited_value|
    @user_metadata.key?(key) ? existing_group_value : new_inherited_value   # <-- line 736
  end
  ...
end
```

`@user_metadata` is only assigned in `ExampleGroup.set_it_up` (called during `subclass`/`describe`). A **singleton example group is never set up that way**, so its `@user_metadata` is `nil`. The merge block only runs on a **key collision**, which is why the example must also carry a key the shared context declares (`:needs_setup` here); with no collision the block never runs and the bug hides.

The result: `nil.key?(key)` → `NoMethodError`. Because this happens inside `Example#run` after the example body, `finished_at` is never set, and `ExecutionResult#calculate_run_time` then raises the misleading `TypeError: can't convert NilClass into an exact number` that users actually see.

Ideally, it would be nice to fix both problems: the metadata inclusion should not trigger an error, and the timing logic should not mask an underlying error via its own `TypeError`.

sjh9714:
## Summary

Fixes #333.

Including a shared group that carries its own metadata into an individual
example raises `NoMethodError: undefined method 'key?' for nil` whenever the
example declares the same metadata key as the shared group. It surfaces as a
confusing `TypeError: can't convert NilClass into an exact number` from
`calculate_run_time`, because the `NoMethodError` escapes after the example
body has run but before timing is recorded.

`ExampleGroup.update_inherited_metadata` resolves key collisions by consulting
`@user_metadata`, which is only assigned in `set_it_up`. The singleton example
group that `Configuration#configure_example` builds for an individual example
never goes through `set_it_up`, so `@user_metadata` is `nil` there. The
`Hash#update` block only runs on a key collision, which is why the crash needs
the same key on both the shared group and the example.

The issue was filed against 3.13.6 with
`shared_context_metadata_behavior = :apply_to_host_groups`. On `main` that
option is gone and the behavior is unconditional, so the reproduction no longer
needs it — the defect is unchanged.

## Changes

- Treat a `nil` `@user_metadata` as "the existing value wins" in
  `ExampleGroup.update_inherited_metadata`. That matches
  `Example#update_inherited_metadata`, which already always keeps the existing
  value; for a singleton example group the metadata is the example's own, so
  its explicitly-set value should take precedence over the inherited one.
- Regression spec beside the existing singleton-class example in
  `configuration_spec.rb`.

I fixed this at the fault site rather than in `with_replaced_metadata`, since
that helper is used elsewhere and receives the example's full metadata rather
than the user-specified subset. Happy to take it the other way if you'd prefer.

## Testing

- `bundle exec rspec spec/rspec/core/configuration_spec.rb -e 'includes a shared example group that has its own metadata in the singleton class of matching examples'`
  — fails on `main` with the `NoMethodError`, passes with the fix.
- `bundle exec rspec spec --seed 1234` — 2155 examples, 18 failures. The same
  18 fail on unmodified `main` (2154 examples, 18 failures); they are all in
  `spec/rspec/core/formatters/exception_presenter_spec.rb` and appear unrelated
  to this change (backtrace formatting under Ruby 4.0.6 locally).
- `bundle exec rubocop rspec-core/lib/rspec/core/example_group.rb rspec-core/spec/rspec/core/configuration_spec.rb`
  — no offenses.

I left the Changelog alone since the entry format wants a PR number; happy to
add one.

Note: I used Claude Code while preparing this change, and I reviewed the final
diff and ran the listed checks locally.

### DIFF
--- a/rspec-core/lib/rspec/core/example_group.rb
+++ b/rspec-core/lib/rspec/core/example_group.rb
@@ -706,7 +706,16 @@ module RSpec
       # @private
       def self.update_inherited_metadata(updates)
         metadata.update(updates) do |key, existing_group_value, new_inherited_value|
-          @user_metadata.key?(key) ? existing_group_value : new_inherited_value
+          # `@user_metadata` is only assigned by `set_it_up`. The singleton
+          # example group used to include a shared group into an individual
+          # example never goes through it, and its metadata is the example's
+          # own, so the existing value wins -- as it does in
+          # `Example#update_inherited_metadata`.
+          if @user_metadata.nil? || @user_metadata.key?(key)
+            existing_group_value
+          else
+            new_inherited_value
+          end
         end
 
         RSpec.configuration.configure_group(self)
--- a/rspec-core/spec/rspec/core/configuration_spec.rb
+++ b/rspec-core/spec/rspec/core/configuration_spec.rb
@@ -1047,6 +1047,20 @@ module RSpec::Core
 
           expect(foo_value).to eq 19
         end
+
+        it 'includes a shared example group that has its own metadata in the singleton class of matching examples' do
+          RSpec.shared_examples "shared group", :shared_group_metadata do
+            let(:foo) { 20 }
+          end
+          RSpec.configuration.include_context "shared group", :include_it
+
+          foo_value = nil
+          describe_successfully do
+            it("", :include_it, :shared_group_metadata) { foo_value = foo }
+          end
+
+          expect(foo_value).to eq 20
+        end
       end
     end
 


pirj (approved):
Looks totally reasonable. Thank you!

Edit: related https://github.com/rspec/rspec-core/pull/2878

PR #339, opened 2026-08-17
Make strict docstrings an opt-in `strict_docstring` option

### LINKED ISSUE #40: doc_string which is the first argument of example, must be string object

pirj:
Supersedes #252 (kept open for now — this branch contains its commits, plus the opt-in on top).

Fixes #40

3.99 counterpart https://github.com/rspec/rspec/pull/107

## Why

#252 makes a non-`String` doc string a hard error in 4.0. Applied unconditionally that breaks every suite that passes a `Symbol` or a `Hash` as the first argument today, which is exactly what @JonRowe pushed back on in review:

> No this is incorrect, this should be interpreted as a `nil` description with metadata, I think that should still be supported for now, just because there are a tonne of aliases and things where that makes sense. Think `describe type: :model { ... }` for example.

The rationale for tightening this up still stands (rspec/rspec-core#3071):

> The publicly documented API is that the first argument is a docstring […] I think I'm leaning towards closing this, and making this a hard requirement of a string in 4.x…

So rather than choose between them, this makes the strictness opt-in, as proposed in https://github.com/rspec/rspec/pull/107#issuecomment-3605985667.

## What

```ruby
RSpec.configure do |config|
  config.strict_docstring! # enforce the very first argument to be a String, to prevent the
                           # confusion caused by an implicit conversion of a Hash/Symbol
                           # meant as metadata into a docstring
end
```

* **Off (the default)** — unchanged RSpec 3 behaviour. `describe :foo` still describes the group as `"foo"` and does not tag it `:foo`; `it(foo: :bar)` still describes the example as `"{foo: :bar}"`. Nobody's suite breaks on upgrade.
* **On** — examples require a `String`, example groups additionally accept a `Class` or a `Module`. Anything else raises `ArgumentError`.

Also:

* Enabled in the `spec/spec_helper.rb` generated by `rspec --init`, so new projects get the stricter behaviour by default, and in this repository's own four suites.
* The example group message now names every accepted type, per @JonRowe's review suggestion: `Example groups must be described with a string, class or module, got: ...`.
* The group overload's `@param doc_string` is widened from `[String]` to `[String, Class, Module]` — it has always accepted all three, the docs just never said so. The example overload's `@param doc_string [String]` was already accurate and is unchanged.

## Specs

`example_group_spec.rb` documents the option as paired contexts, so the difference is legible side by side:

```ruby
context "when `strict_docstring` is not enabled" do
  it "treats the first argument as the description when it is a hash" do
    group = RSpec.describe(:foo => :bar)

    expect(group.description).to eq({ :foo => :bar }.to_s)
    expect(group.metadata).not_to include(:foo)
  end
end

context "when `strict_docstring` is enabled" do
  before { RSpec.configuration.strict_docstring! }

  it "raises ArgumentError when a Hash is used as an example group doc string" do
    expect { RSpec.describe(:foo => :bar) { } }.to raise_error(ArgumentError, /…/)
  end
end
```

The specs #252 removed (`does not treat the first argument as a metadata key even if it is a symbol`, `treats the first argument as part of the description when it is a symbol`, and `metadata_spec.rb`'s "with a Symbol") are back, since that behaviour is live again by default.

`features/configuration/strict_docstring.feature` covers it end to end. Its first scenario demonstrates the confusion the option prevents: `it :slow do` passes and prints "slow" under `--format doc`, then `--tag slow` finds 0 examples.

## Not addressed here

@JonRowe's review also suggests a leading `Symbol`/`Hash` should be *interpreted as metadata* (`describe type: :model` → a `nil` description tagged `type: :model`). Making the check opt-in removes the breakage without requiring that call; interpreting metadata is a behaviour change in its own right and is left as a separate decision.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JonRowe:
Please don't add copilot reviews, or add Claude commits. Our policy is to declare AI use (and I reserve the right to reject a PR if I think its just slop) but only a human should be credited as they are the ones responsible for the code committed.

pirj:
Re-checked every single line of this, and confirm that I'm fully responsible for this.

I'll fix RuboCop in a separate PR.

PR #340, opened 2026-08-17
Declare Ruby 4.0 bundled gems in the Gemfile

pirj:
## The problem

Ruby 4.0 promotes a batch of default gems to **bundled** gems. A bundled gem ships with Ruby but is not requirable unless the Gemfile declares it, so anything that requires one dies under `bundle exec`.

`aruba` requires `logger` from `Aruba::ArubaLogger`, so on Ruby 4.0 every suite that loads aruba fails at `spec_helper`, before a single example runs:

```
An error occurred while loading spec_helper.
Failure/Error: require 'aruba/api'

LoadError:
  cannot load such file -- logger
# aruba-2.4.1/lib/aruba/aruba_logger.rb:3:in '<top (required)>'
# aruba-2.4.1/lib/aruba/platforms/unix_platform.rb:7:in '<top (required)>'
# aruba-2.4.1/lib/aruba/platform.rb:3:in '<top (required)>'
# aruba-2.4.1/lib/aruba/api.rb:8:in '<top (required)>'
# ./spec/support/aruba_support.rb:7:in 'block in <module:ArubaLoader>'
```

`rspec-core` and `rspec-mocks` both report `0 examples, 0 failures, 1 error occurred outside of examples`. Cucumber dies the same way, straight out of Ruby's own loader:

```
ruby/4.0.0/bundled_gems.rb:60:in 'Kernel.require': cannot load such file -- logger (LoadError)
```

## The precise cut

Ruby's own `Gem::BUNDLED_GEMS::SINCE` table is the authority on when each gem changed status (`ruby/4.0.0/bundled_gems.rb`):

```ruby
"bigdecimal" => "3.4.0",
"drb"        => "3.4.0",
...
"logger"     => "4.0.0",
"benchmark"  => "4.0.0",
"ostruct"    => "4.0.0",
...
"tsort"      => "4.1.0",
```

`RUBY_VERSION < SINCE[name]` is a warning; at or above it, a hard `LoadError`. So the cut for `logger` is exactly `4.0.0`, and the fix belongs in the Gemfile's existing "Support for gems extracted in Ruby versions" section alongside `bigdecimal` and `drb`, gated the same way.

## Why this PR opens with an empty commit

The first commit is deliberately empty, so this PR's first CI run is `main` unchanged. That gives a red baseline on the Ruby 4.0 matrix entries and shows exactly which bundled gems actually bite across all four libraries — rather than guessing at the list and declaring gems we do not need. The fix goes on top of that evidence, and the PR comes out of draft once the matrix is green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pirj:
Closing: the premise did not hold.

All 44 checks pass on this empty commit against unmodified `main`, including every `Ruby 4.0` job (core, expectations, mocks, support) and jruby-10.1. There is no bundled-gem problem in the repo to fix.

The `LoadError: cannot load such file -- logger` I was chasing was purely a local environment artifact: bundler 2.6.2 driving RubyGems 4.0.3. CI sidesteps this by running `script/update_rubygems_and_install_bundler` (which installs a bundler matched to the current RubyGems) before `script/bundle`. Upgrading locally to RubyGems/bundler 4.0.13 resolves it.

Ruby's `Gem::BUNDLED_GEMS::SINCE` table does list `"logger" => "4.0.0"`, so the hazard is real in principle — but it is not biting the actual CI matrix, and declaring the gem speculatively is not worth it.

